PR #1, opened 2013-01-25
rosbuild fails when building in isolation

dirk-thomas:
When building in isolation the packages are not in the same hierarchy as in the repo. This makes several references inside the compatibility layer fail.

Most prominent each dry package expects this file:
  ${ROS_ROOT}/core/rosbuild/rosbuild.cmake


tfoote:
This looks correct.  We reviewed all files installed into ROS_ROOT from fuerte and all are accounted for.

Issue #4, opened 2013-02-19
rosmake: use -jN -lN for parallel builds


dirk-thomas:
rosmake should use both argument with N (similar as ros/catkin#326).

dirk-thomas:
Consider backporting a6375042f362e0bbb46f8c99f939873ba928bd99 and 4a4f6b9db302a4d19e2139e2688cddf1138240e9 to Fuerte.

tkruse:
How about also patching electric and releasing electric? See 4db5ac6a9fac4fec

dirk-thomas:
Since electric is officially eol / not supported anymore I am not planning to backport the patch to electric.

tkruse:
Thanks you very much for this explanation. Knowing that the excellent ROS support has in the past also provided critical patch releases beyond the eol of distributions, I thought this might be an instance where this could also be applied.
If this were not to happen, I would most humbly suggest that the electric users, which according to this recent poll are still numerous:
https://docs.google.com/spreadsheet/viewanalytics?formkey=dFpudUFPTW5GQ0dES1MyZjlUbHN6OHc6MQ
would be informed about an alternative. I can offer to provide that information myself using the existing communication channels for announcements if that is any help at all.

EDIT: Rephrased to sound more positive and constructive.

tfoote:
@tkruse Electric is now the 3rd oldest released distro.  We have historically provided support for two versions of ROS at any given time.  We do still provide the packages for downloads and will continue to provide critical patch releases assuming the buildfarm is still operational.

tkruse:
@tfoote: Can you please kindly state whether this issue is a candidate for a critical patch that way or not.

Adding the patch to the tip of an electric branch without a release may still help electric users building from source.

tfoote:
If you want to submit a pull request we certainly won't turn that down.  I
don't think that this is critical enough for us to rebuild the entirity of
electric.  There's been no user complaints and it doesn't block any
development.

tkruse:
I created a pull request indeed:
https://github.com/ros/ros/pull/5

I can accept that if the effort of rebuilding electric is high, you cannot reasonably provide a new electric release. 
In that case I would like to see a simple message to electric users added to the release notes for the next update of fuerte + groovy, that explains how to improve their ROS Electric experience.

Can you give me an estimate of when this patch will be released for fuerte and groovy?

PR #5, opened 2013-02-27
rosmake use -jN -lN (backported a6375042f362e0bbb46f8c99f939873ba928bd99...

tkruse:
... and 4a4f6b9db302a4d19e2139e2688cddf1138240e9, #4)


dirk-thomas:
I saw that this was also the case for groovy and fuerte - I just fixed that.

PR #6, opened 2013-03-10
Path manipulation for windows

stonier:
The backtick command wasn't working on 64 bit machines due to the brackets in path directories such as `C:\Program Files (x86)\Microsoft ...`. It would drop out with the amusing error:

```
\Microsoft was unexpected at this time
```

This fixes that.


dirk-thomas:
Can you use a variable with a prefix like `_` to not collide with a potential user space variable called `COMMAND`?

stonier:
_COMMAND works fine.

PR #14, opened 2013-05-08
make eclipse-project issues

smits:
In groovy the make eclipse-project target is broken, a first commit adds the inclusion of some missing environment variables that are needed to fix building from within eclipse, a second commit adds some cleanup of catkin generated files and directories.

Ruben

### DIFF
--- a/core/mk/cmake.mk
+++ b/core/mk/cmake.mk
@@ -46,6 +46,6 @@ eclipse-project:
 	awk -f $(shell rospack find mk)/eclipse.awk .project-cmake > .project
 	rm .project-cmake
 	python $(shell rospack find mk)/make_pydev_project.py
-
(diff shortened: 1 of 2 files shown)


tfoote:
I'm not familiare with eclipse-project but it looks reasonable +1

dirk-thomas:
@smits I don't understand why your change 25643a2f93e84eecd7da91d0fedc34f0146c10e3 removes those files and folders. They will be generated during the next make cycle anyway. Can you comment on this? Else I might just revert this.

dirk-thomas:
This is really nasty. The build folder of the CMake project is basically moved after generating the Eclipse project.

Since dry does not support out-of-source builds cleanly and Eclipse CDT projects should not be created in subfolders of the source there seems to be not much that we can do about that...

PR #16, opened 2013-05-31
Added shell functions for fish shell

aclough:
Hello, this is a commit to provide a version of the various shell functions found in ros bash to the fish shell.  Specifically, this is for the forked version under current development on github rather than the legacy version in the Debian/Ubuntu repositories.  I don't know anyone currently using that version.

I've duplicated all the functions, but none of the auto completions in this patch.  rospd works slightly differently but mirrors the way in which fish does pushing directories.

More info about fish:
https://en.wikipedia.org/wiki/Fish_shell
https://github.com/fish-shell/fish-shell

### DIFF
--- a/tools/rosbash/rosfish
+++ b/tools/rosbash/rosfish
@@ -1,0 +1,278 @@
+function _rossed
+  if test (uname) = "Darwin" -o (uname) = "FreeBSD"
+      sed -E $argv
+  else
+      sed -r $argv
+  end
+end
+
+function _rosfind
+  if  test (uname) = "Darwin" -o (uanme) = "FreeBSD"
+    # BSD find needs -E for extended regexp
+    find -E $argv
+  else
+    find $argv
+  end
+end
+
+# _ros_location_find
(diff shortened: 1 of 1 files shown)


dirk-thomas:
When do you expect the official Deban/Ubuntu packages be released which work with the added scripts?

The pull request needs some more changes that your scripts gets actually usable for users:
- your file must be installed else it is not available in the rosbash Debian package, see https://github.com/ros/ros/blob/groovy-devel/tools/rosbash/CMakeLists.txt#L6
- furthermore for all other shells there are environment hooks in place which source the specific files, see https://github.com/ros/ros/tree/groovy-devel/tools/rosbash/env-hooks

Since fish is currently not supported by ROS and catkin (which generated the main environment files setup.\* under /opt/ros/DISTRO) this would also require a fish script in order to source your fish-specific environment hooks.

aclough:
Ok, I'll start work on some commits dealing with those issues.

dirk-thomas:
Closing due to lack of response.

PR #19, opened 2013-06-20
enable roslaunch args completion in rosbash

dornhege:
Depends on [#230 in ros_comm](https://github.com/ros/ros_comm/pull/230) to work. Will use default completion if that is not available.

### DIFF
--- a/tools/rosbash/rosbash
+++ b/tools/rosbash/rosbash
@@ -479,12 +479,26 @@ function _roscomplete_launch {
     arg="${COMP_WORDS[COMP_CWORD]}"
     COMPREPLY=()
     if [[ ${arg} =~ \-\-.* ]]; then
-        COMPREPLY=(${COMPREPLY[@]} $(compgen -W "--files --args --nodes --find-node --child --local --screen --server_uri --run_id --wait --port --core --pid --dump-params" -- ${arg}))
+        COMPREPLY=(${COMPREPLY[@]} $(compgen -W "--files --args --nodes --find-node --child --local --screen --server_uri --run_id --wait --port --core --pid --dump-params --skip-log-check --ros-args" -- ${arg}))
         
     else
         _roscomplete_search_dir "( -type f -regex .*\.launch$ -o -type f -regex .*\.test$ )"
         if [[ $COMP_CWORD == 1 ]]; then
            COMPREPLY=($(compgen -o plusdirs -f -X "!*.launch" -- ${arg}) ${COMPREPLY[@]} $(compgen -o plusdirs -f -X "!*.test" -- ${arg}) ${COMPREPLY[@]})
+        fi
+        # complete roslaunch arguments for a launch file
+        if [[ ${#COMP_WORDS[@]} -ge 2 ]]; then
+            ROSLAUNCH_COMPLETE=$(which roslaunch-complete)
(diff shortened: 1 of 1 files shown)


tools/rosbash/rosbash
```diff
@@ -479,13 +479,27 @@ function _roscomplete_launch {
...
-        COMPREPLY=(${COMPREPLY[@]} $(compgen -W "--files --args --nodes --find-node --child --local --screen --server_uri --run_id --wait --port --core --pid --dump-params" -- ${arg}))
+        COMPREPLY=(${COMPREPLY[@]} $(compgen -W "--files --args --nodes --find-node --child --local --screen --server_uri --run_id --wait --port --core --pid --dump-params --skip-log-check --ros-args" -- ${arg}))
         
     else
         _roscomplete_search_dir "( -type f -regex .*\.launch$ -o -type f -regex .*\.test$ )"
         if [[ $COMP_CWORD == 1 ]]; then
            COMPREPLY=($(compgen -o plusdirs -f -X "!*.launch" -- ${arg}) ${COMPREPLY[@]} $(compgen -o plusdirs -f -X "!*.test" -- ${arg}) ${COMPREPLY[@]})
         fi
+        # complete roslaunch arguments for a launch file
+        if [[ ${#COMP_WORDS[@]} -ge 2 ]]; then
```
dirk-thomas, line 490: The logic seems to not work when passing a launch file to roslaunch (instead of pgkname and launch file name).

tools/rosbash/rosbash
```diff
@@ -479,13 +479,27 @@ function _roscomplete_launch {
...
+            ROSLAUNCH_COMPLETE=$(which roslaunch-complete)
+            if [[ -x ${ROSLAUNCH_COMPLETE} ]]; then
+                # Call roslaunch-complete instead of roslaunch to get arg completion
+                _roslaunch_args=$(${ROSLAUNCH_COMPLETE} ${COMP_WORDS[@]:1:2})
```
dirk-thomas, line 494: Same as 490

tools/rosbash/rosbash
dornhege, line 490: This should work. We will need at least two full words to hand over for args completion: roslaunch itself + the launch file, e.g. `roslaunch x.launch`. Adding more like `roslaunch pkg x.launch` won't hurt.

tools/rosbash/rosbash
dornhege, line 494: This will pass the first two arguments of roslaunch to roslaunch-complete. This is all we need. If it is only the launch file, only that will be passed. If it is pkg + launch, both will be passed. Any additional parameters like `roslaunch x.launch ARG` won't hurt.

The idea is that at this point we can't easily figure out what is intended. Someone might name the launch file `robot_bringup` (no .launch) or similar. This logic is left to roslaunch-complete to figure out in the same way that roslaunch does that. `rlutil.resolve_launch_arguments(sys.argv[1:])` is used in there to deal with the two possibilities of calling roslaunch. This makes sure we always behave like roslaunch would.

dirk-thomas, line 494: Can you please update this line to ignore output to stderr:

```
_roslaunch_args=$(${ROSLAUNCH_COMPLETE} ${COMP_WORDS[@]:1:2} 2> /dev/null)
```

PR #25, opened 2013-07-23
Gendeps cache fix

tkruse:
This bug request follows a report of slow performanceof gendeps from Anthony Mallet at LAAS, for building ROS packages over NFS.

It seems to me the gendeps script is not using the rospack instance caching as intended by rospack and gentools.py.

To make sure this improves performance, I intentionally made my ROS_PACKAGE_PATH include large non-ROS subtrees to simulate lenghty file IO, such as 

```
 export ROS_PACKAGE_PATH:$ROS_PACKAGE_PATH:/home/kruset
```

Then for testing, I used this call in a groovy underlay from source:

```
$ time install/lib/roslib/gendeps --md5 src/nav_msgs/msg/OccupancyGrid.msg 
3381f2d731d4076ec5c71b0759edbe4e
real    0m10.908s
user    0m6.404s
sys     0m4.400s
```

(after calling several times to initialize rospack cache)
After applying the path, i get results such as this:

```
real    0m2.960s
user    0m1.888s
sys     0m1.032s
```

MD5 calculationthen is still slow, will investigate further.

If possible, please apply to fuerte, groovy and hydro sources (even if fuerte won't be released with patch).

Note that with all the duplicate code of md5 calculation between roslib and genpy/genmsg, I cannot tell whether the latter suffers the same performance hits. If this PR is considered valid, I recommend checking in genpy/genmsg as well.

### DIFF
--- a/core/roslib/scripts/gendeps
+++ b/core/roslib/scripts/gendeps
@@ -87,7 +87,7 @@ def gendeps_main(argv, stdout, stderr):
 
     # rospack instance for caching of deps
     rospack = rospkg.RosPack()
-    retval = roslib.gentools.get_file_dependencies(f, stdout=stdout, stderr=stderr)
+    retval = roslib.gentools.get_file_dependencies(f, stdout=stdout, stderr=stderr, rospack=rospack)
 
     if options.md5:
         print >> stdout, roslib.gentools.compute_md5(retval, rospack=rospack)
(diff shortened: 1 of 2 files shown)


dirk-thomas:
Thank you, the patch looks good. If without running it I can see the improvement.

This will only affect dry message generation. The wet counterpart does not rely on `RosPack` at all.

dirk-thomas:
Merged to branches: ros-1.8 (fuerte), groovy-devel and hydro-devel.

PR #26, opened 2013-07-24
lazy initialize valid package names

tkruse:
This can save rospack roscache update call, when only primitive types are used (E.g. Pose2D below uses 3 floats).

Again, with a ROS_PACKAGE_PATH that includes a lot of useless stuff, I tried without and with the patch:

Before:

```
$ time install/lib/roslib/gendeps src/geometry_msgs/msg/Pose2D.msg 
real    0m3.348s
user    0m2.172s
sys     0m1.160s
```

After:

```
$ time install/lib/roslib/gendeps --md5 src/geometry_msgs/msg/Pose2D.msg 
938fa65709584ad8e77d238529be13b8
real    0m1.119s
user    0m0.728s
sys     0m0.388s
```

I think the patch also makes sense by just reading it. The variable valid_packages is initialized even when not used, and initialization involves rospack cache update.

### DIFF
--- a/core/roslib/src/roslib/gentools.py
+++ b/core/roslib/src/roslib/gentools.py
@@ -70,14 +70,18 @@ def _add_msgs_depends(rospack, spec, deps, package_context):
     @type  deps: [str]
     @raise KeyError for invalid dependent types due to missing package dependencies.
     """
-    valid_packages = ['', package_context]
-    try:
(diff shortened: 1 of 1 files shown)


tkruse:
Again, if valid, please apply to fuerte and hydro as well (in source at least, even if no new release).

PR #27, opened 2013-08-02
replaced \t with tab character

muratsevim:
The sed installed on OSX Lion treats \t as a regular 't' char which raises issues if there is a 't' char in home path.

### DIFF
--- a/tools/rosbash/rosbash
+++ b/tools/rosbash/rosbash
@@ -27,7 +27,8 @@ function _rosfind {
 # stack echoes its result
 function _ros_location_find {
     local homedir ROS_LOCATION_KEYS_ARR ROS_LOCATIONS_ARR loc
-    homedir=`echo $HOME | sed -e "s/\//\t\//g" -e "s/\t/\\\\\/g"`
+    TAB=$'\t' 
+    homedir=`echo $HOME | sed -e "s/\//${TAB}\//g" -e "s/${TAB}/\\\\\/g"`
     ROS_LOCATION_KEYS_ARR=(`echo $ROS_LOCATIONS | _rossed -e 's/([^:=]*)=([^:=]*)(:*[^=])*(:|$)/\1 /g'`)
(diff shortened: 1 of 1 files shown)


dirk-thomas:
@wjwwood Can you confirm the issue before the patch and that it works with the patch applied?

wjwwood:
Yes, I'll test that now.

wjwwood:
@muratsevim what do you mean by home path? A package of that name in the ROS_PACKAGE_PATH?

wjwwood:
Nvm, I see from the patch, if the home directory has a `t` in it, does it need to be at the front or anywhere in it?

wjwwood:
I can reproduce the problem and the fix:

``` bash
$ export HOME=/tmp/rosbash_ws/thome

$ homedir=`echo $HOME | sed -e "s/\//\t\//g" -e "s/\t/\\\\\/g"`

$ echo $homedir
\/\mp\/rosbash_ws\/\home

$ export HOME=/tmp/rosbash_ws/thome

$ TAB=$'\t'

$ homedir=`echo $HOME | sed -e "s/\//${TAB}\//g" -e "s/${TAB}/\\\\\/g"`

$ echo $homedir
\/tmp\/rosbash_ws\/thome
```

dirk-thomas:
The goal here is to replace `~` with `$HOME`. Using `/` as the character to separate the reg exp parts is probably the worst decision in the rosbash code. By using `|` as the delimiter no escaping of `$HOME` would be necessary in the first place.

dirk-thomas:
I have created a modified pull request #28. @muratsevim Can you please verify that this also fixes your issue?

muratsevim:
@dirk-thomas I have tested #28 and can confirm it fixes my issue.

For your information, I don't actually use `~` in any of my ros paths but this was causing errors for me because the last `t` char on my home path was escaping the next `/` on the sed expression.

dirk-thomas:
Thanks for checking and providing the patch.

PR #29, opened 2013-08-03
resolve rosunit path in installspace at runtime without find_program(), update env hooks to use runtime context (ros/catkin#490)

dirk-thomas:


tfoote:
+1

PR #30, opened 2013-08-23
Show error message and return code when invalid path passed to rospd.

mikepurvis:
Increases the scriptability of rospd to have it return 1 (like roscd) when the package can't be found.

### DIFF
--- a/tools/rosbash/rosbash
+++ b/tools/rosbash/rosbash
@@ -210,9 +210,14 @@ function rospd {
     else
         local rosvals
         _ros_decode_path $1 forceeval
+        if [ $? != 0 ]; then
+            echo "rospd: No such package/stack '$1'"
+            return 1
+        fi
         pushd ${rosvals[1]}${rosvals[2]}${rosvals[3]} > /dev/null ;
     fi
     rosd
+    return 0
 }
 
 function rosls {


mikepurvis:
Now that hydro is released, would it make more sense to provide a PR for hydro instead?

dirk-thomas:
If you are fine with this only be addressed in Hydro you could provide a pull request against Hydro instead of Groovy. But if it should be changed for both this pull request is fine - it would be cherry-picked for Hydro after being merged.

Could you describe briefly some invocations you tried to check the behaviour (that would make it easier for me to validate the changes - besides reading the diff)? I will look at this in a couple of days.

mikepurvis:
Of course:

``` bash
source /opt/ros/groovy/setup.bash
source ~/ros/tools/rosbash/rosbash
rospd roscpp  # *
echo $?    # expect 0
rospd foo  # expect "No such package"
echo $?    # expect 1
rosd       # expect same output as *
popd       # expect to return to where you started.
```

dirk-thomas:
Your pull request breaks the behavior of invoking `rospd` without arguments. Currently it toggles between the last two folders. With your patch it says "No such package".

I created a fixed pull request: #38. Can you please confirm that this one works for you too?

Issue #2, opened 2013-01-31
rosmake: add option to disable rosdep check in rosmake  (ros ticket #3928)


tfoote:
To support a work around for #3927

A quick fix is to comment out the call to rosdep.

trac data:
- Owner: **tfoote**
- Reporter: **tfoote**
- Reported at: **Tue Apr 17 15:14:05 2012**
- URL: https://code.ros.org/trac/ros/ticket/3928

tfoote:
[joq] Seems like a good idea. 

Building source rosdeps on a machine without network access is a legitimate need.

dirk-thomas:
Related to ros/rosdep#197

tfoote:
We've checked on both fuerte and groovy that rosmake does not call rosdep.

PR #32, opened 2013-08-30
Added shell functions for fish shell #2

aclough:
Continuation of this pull request: https://github.com/ros/ros/pull/16

Sorry for taking so long to get back to this, there was a lot of work to get done on other things before heading off on vacation and I'm afraid this fell to the wayside.  I think this addresses all the concerns you had.

The version of fish this is made for is finally going into mainline Ubuntu with 13.10.
http://packages.ubuntu.com/saucy/fish

### DIFF
--- a/tools/rosbash/CMakeLists.txt
+++ b/tools/rosbash/CMakeLists.txt
@@ -3,7 +3,7 @@ project(rosbash)
 find_package(catkin)
 catkin_package()
 
-install(FILES rosbash rostcsh roszsh
+install(FILES rosbash rostcsh roszsh rosfish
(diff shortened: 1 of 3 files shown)


dirk-thomas:
Since non of the current ROS distros is built for Ubuntu 13.10 and the functionality would not work on the other distos this can not be easily merged. Probably it has to wait (quite some time) until Ubuntu catches up.

Alternatively you could create a separate package _rosfish_ which interested users could checkout and build fro source?

aclough:
That's fair.  I'll keep this up to date for my own use in the mean time and push this again when things are better synced up.

PR #35, opened 2013-10-02
fix sanitizing rosunit xml files on the lowest level possible

dirk-thomas:
@wjwwood @tfoote Please review.

Should also be backported to groovy-devel.


tfoote:
It would be cleaner in my mind to put the write back inside the sanitizing method since it's working around the same issue.  And possibly add an argument write_back_sanitized=True to the prototype for clarity. And possibly consider renaming the function, though that's not strictly necessary if the parameter is added. 

But it should get the job done.

tfoote:
+1

PR #38, opened 2013-10-03
fix return code for rospd for invalid package names (fix #30)

dirk-thomas:
@mikepurvis @wjwwood Please review

Also cherry-pick to Groovy after merge.


mikepurvis:
LGTM

```
$ source /opt/ros/hydro/setup.bash
$ source ~/ros/tools/rosbash/rosbash
$ rospd sensor_msgs
$ rospd
$ rospd foo
$ echo $?
1
```

dirk-thomas:
Thank you for verifying that it works for you.

PR #42, opened 2013-12-05
python 3 compatibility

dirk-thomas:


tfoote:
+1

core/roslib/test/test_roslib_manifest.py
```diff
@@ -143,8 +143,8 @@ def test_parse_bad_file(self):
       try:
         parse_file(p)
         self.fail("parse should have failed on bad manifest")
-      except ManifestException, e:
-        print str(e)
+      except ManifestException as e:
+        print(str(e))
```
wjwwood, line 147: Is it the case that `'{0}'.format(e)` would work better in Python2 if the exception contains unicode?

wjwwood, line 147: Actually I guess not unless you imported `unicode_literals` from `__future__`.

wjwwood:
+1

dirk-thomas:
Superseded by #45.

PR #56, opened 2014-06-03
use PYTHON_EXECUTABLE in environment hooks

dirk-thomas:
This requires ros/catkin#645.

@wjwwood Please review.


wjwwood:
Seems fine to me, just so long as `PYTHON_EXECUTABLE` is always available at templating time. How is that variable provided, by `roslib` or by something else?

wjwwood:
Nvm, I see that it is provided by ros/catkin#645

wjwwood:
+1

Issue #44, opened 2014-02-05
warning with cmake 2.8.12.2: Argument not separated from preceding token by whitespace


dirk-thomas:
Filled by @dpinol at ros/catkin#586:

When compiling a dry package with cmake 2.8.12.2, I get this warning
You can try it with: https://github.com/dpinol/ros_tests/tree/master/cmake_policy

The fix consists in quoting the second argument of the add_custom_target call (escaping the echo's quotes)

```
~/svn/catkin_ws/ros_tests/cmake_policy$ make
mkdir -p bin
cd build && cmake -Wdev -DCMAKE_TOOLCHAIN_FILE=/opt/ros/hydro/share/ros/core/rosbuild/rostoolchain.cmake  ..
-- The C compiler identification is GNU 4.6.3
-- The CXX compiler identification is GNU 4.6.3
-- Check for working C compiler: /home/dani/bin/cc
-- Check for working C compiler: /home/dani/bin/cc -- works
-- Detecting C compiler ABI info
-- Detecting C compiler ABI info - done
-- Check for working CXX compiler: /home/dani/branches_svn/scm/bin/utils/developer_utils/g++
-- Check for working CXX compiler: /home/dani/branches_svn/scm/bin/utils/developer_utils/g++ -- works
-- Detecting CXX compiler ABI info
-- Detecting CXX compiler ABI info - done
-- Found PythonInterp: /usr/bin/python (found version "2.7.3") 
CMake Warning (dev) at /opt/ros/hydro/share/ros/core/rosbuild/rosbuild.cmake:40 (include):
  Syntax Warning in cmake code at

    /opt/ros/hydro/share/ros/core/rosbuild/public.cmake:330:140

  Argument not separated from preceding token by whitespace.
Call Stack (most recent call first):
  CMakeLists.txt:2 (include)
This warning is for project developers.  Use -Wno-dev to suppress it.

[rosbuild] Building package cmake_policy
```

PR #60, opened 2014-07-02
Explicit disabling of batch file delayed expansion

JonathanWade:
Windows environment hooks are a nesting of batch files. Changes in the way catkin sets up environment hooks will likely require the use of delayed expansion in upper level batch scripts that call this script. Delayed expansion settings are inherited through the chain of execution when batch files call each other.

Although it was commented that delayed expansion should be disabled for the period of echo'ing python code, this patch explicitly disables delayed expansion for that portion of the file.

See this pull request in win_ros for further context: [win_ros pull request #44](https://github.com/ros-windows/win_ros/pull/44)

### DIFF
--- a/core/roslib/env-hooks/10.ros.bat.em
+++ b/core/roslib/env-hooks/10.ros.bat.em
@@ -9,6 +9,7 @@ set ROS_DISTRO=hydro
 
 REM python function to generate ROS package path based on all parent workspaces (prepends the separator if necessary)
 REM do not use EnableDelayedExpansion here, it messes with the != symbols
+setlocal disabledelayedexpansion
 echo from __future__ import print_function > _parent_package_path.py
 echo import os >> _parent_package_path.py
 echo env_name = 'CATKIN_WORKSPACES' >> _parent_package_path.py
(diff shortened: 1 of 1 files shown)


dirk-thomas:
You mentioned "likely require". Did you actually ran into a problem, made the proposed change and then confirmed that it fixed the issue?

JonathanWade:
Yes, this is one piece of a three file solution to the new catkin env_hooks style for use in windows. The top level change is [here](https://github.com/ros/catkin/pull/673). As you can see that top setup.bat requires the use of delayed expansion to loop through the environment hook variables. This fix is required or else the `+=` operators in the python code become mangled. While the setup.bat pull may require more vetting or some additional thought towards a more graceful solution, this change is benign and should guard against accidental code mangling in the future even if the setup.bat changes go unused.

dirk-thomas:
You mentioned a "three file solution". What is the third pieces beside this patch and the referenced catkin ticket?

JonathanWade:
The third piece is specific to windows-ros' fork of ros_comm: https://github.com/ros-windows/ros_comm/pull/2

dirk-thomas:
I have cherry-picked the patch to the _indigo-devel_ branch: 591df402d26585a288f6774f9380188445d984a7

I will be considered for backporting before the next Hydro patch release.

dirk-thomas:
The third piece has also been cherry-picked to the _indigo-devel_ branch of `ros_comm`: ros/ros_comm@a735f6456bd64baf24e891993161a2fbad5551ad

Issue #62, opened 2014-07-23
rosboost_cfg can't find boost::system on Indigo


trainman419:
As seen on ROS Answers: http://answers.ros.org/question/187492/xv_11_laser_driver-build-error/

```
CMake Error at /opt/ros/indigo/share/ros/core/rosbuild/public.cmake:847 (message):
  [rosboost-cfg --libs system] failed with error: Traceback (most recent call
  last):

    File "/opt/ros/indigo/bin/rosboost-cfg", line 35, in <module>
      rosboost_cfg.main()
    File "/opt/ros/indigo/lib/python2.7/dist-packages/rosboost_cfg/rosboost_cfg.py", line 359, in main
      output = lflags(ver, options.lflags.split(','))
    File "/opt/ros/indigo/lib/python2.7/dist-packages/rosboost_cfg/rosboost_cfg.py", line 278, in lflags
      s += lib_flags(ver, lib) + " "
    File "/opt/ros/indigo/lib/python2.7/dist-packages/rosboost_cfg/rosboost_cfg.py", line 268, in lib_flags
      lib = find_lib(ver, name)
    File "/opt/ros/indigo/lib/python2.7/dist-packages/rosboost_cfg/rosboost_cfg.py", line 238, in find_lib
      raise BoostError('Could not locate library [%s], version %s'%(name, ver))

  rosboost_cfg.rosboost_cfg.BoostError: "Could not locate library [system],
  version (1, 54, 0, '/usr', '/usr/include', True, True)"

Call Stack (most recent call first):
  CMakeLists.txt:29 (rosbuild_link_boost)
```

dirk-thomas:
Under Saucy / Trusty Boost libraries are installed to `/usr/lib/x86_64-linux-gnu` and are therefore not found by the dry rosboost_cfg script.

dirk-thomas:
Fix released in ros-indigo-ros 1.11.4.

PR #63, opened 2014-07-23
fix dry rosboost_cfg to support finding boost in multiarch lib folder

dirk-thomas:
Addresses #62.

@tfoote @wjwwood Please review.


wjwwood:
+1, how annoying.

tfoote:
+1

Issue #48, opened 2014-03-27
rosbash: zsh auto-completion issues


jbohren:
- [x] roslaunch: good
- [ ] rosrun: 
  - package auto-completes
  - executable **does not autocomplete for installed executables**
  - executable **autocompletes git hooks in addition to executables**
- [x] roscd: good
- [ ] rostopic pub:
  - topic name auto-completes
  - message type auto-completes (but it would be nice if this was determined automatically if the topic is already subscribed to)
  - message cmd-line YAML **does not autocomplete**

dirk-thomas:
Thanks for summarizing the state that detailed. Please consider to provide patches for improved `zsh` support. Since most people are only using `bash` this won't be high on the priority list.

dirk-thomas:
@wjwwood since you use zsh on Mac you might want to take a look at this at some point.

trainman419:
Tab-completion for installed executables in rosrun should be fixed in https://github.com/ros/ros/commit/bba94288c13e8c0d0b433ccadfa4b12249838659

bajo:
@jbohren Could you shed a little light on the issue that rosrun is not able to autocomplete installed executables? Can you try a current version and report if this still is a problem, because I think @trainman419 resolved this in https://github.com/ros/ros/commit/bba94288c13e8c0d0b433ccadfa4b12249838659 

If this is the case you could update and close this issue as this and the rostopic autocompletion  https://github.com/ros/ros/commit/dc18b0042afbcfaf9d13d41ca2d15e3f96934839 are working now.

Thanks.

dirk-thomas:
It seems to also work for me. Therefore closing the ticket for now. If there is more information about what is not working please feel free to comment and it can be reopened.

furushchev:
Was this issue solved? I still doubt it does work on zsh when `rosrun` and `rostopic`.

bajo:
@furushchev Can you describe what exactly is not working for you?
I just tested it with rostopic pub <Tab> <Tab> <Tab> and it completed it to 
rostopic pub /rosout rosgraph_msgs/Log "header:
  seq: 0
  stamp: {secs: 0, nsecs: 0}
  frame_id: ''
level: 0
name: ''
msg: ''
file: ''
function: ''
line: 0
topics: ['']"

rosrun rv<Tab> <Tab> autocompletes to rosrun rviz rviz. Just as expected.
[...]

PR #64, opened 2014-07-30
rostopic autocompletion for published topics, msg-type and YAML

bajo:
This should fix parts of issue #48 rosbash: zsh auto-completion issues

### DIFF
--- a/tools/rosbash/roszsh
+++ b/tools/rosbash/roszsh
@@ -470,7 +470,7 @@ function _roscomplete_rosparam {
 
 }
 function _roscomplete_rostopic {
-    local opts
+    local opts word
     reply=()
 
     if [[ ${CURRENT} == 2 ]]; then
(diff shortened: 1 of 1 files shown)


dirk-thomas:
@wjwwood can you please try this on OSX?

wjwwood:
Working on it.

wjwwood:
@bajo what is the expected behavior before and after this patch? I am able to tab complete on zsh on my mac, but I'm not sure what exactly I am looking for.

wjwwood:
I will say the tab completion does look better, I just wanted a more exact list of things you expect to be fixed by this.

bajo:
@wjwwood Before the patch, at least on my machine, the tab completion did not work for the topic name, the msg-type and the data in yaml format. If you want to send a message with a type like geometry_msgs/PoseWithCovarianceStamped you need to type all the fields and the data by hand and watch out for problems with the format and escaping of data, which is kind of annoying. :)

After the patch the tab completion should be able to complete the topic name, based on the already available topics. The type of the topic is retrieved and completed, so no std_msgs/String publishing on a topic which expects something like a geometry_msgs/PoseStamped and the prototype of the message is also completed. In the prototype you can just fill in the parts of data you want to publish.

dirk-thomas:
Thanks for clarifying. The patch looks good.

There is one more command completion in #48 which does not yet work completely in zsh: `rosrun` Could I convince you to also provide a pull request for that? :wink:

wjwwood:
+1 from me, that is what I observed, but I just wanted to make sure you didn't fix anything else I should test too.

Thanks for this!

bajo:
I'll have a look at that problem as well, after my weekend trip. ;)

PR #67, opened 2014-10-15
"std_msgs/Header header" is not considered to be a header  in messages generated with rosbuild

dirk-thomas:
Follow up of ros/gencpp#21.

The message parser used for `rosbuild` only considers the field `Header header` to be a header.

`std_msgs/Header header` is not accepted which leads to not generating certain message traits: https://github.com/ros/ros/blob/01f9fc1140e2940be2799688aca85d5441807fa7/core/roslib/src/roslib/msgs.py#L304

### DIFF
--- a/core/roslib/src/roslib/msgs.py
+++ b/core/roslib/src/roslib/msgs.py
@@ -301,7 +301,7 @@ class MsgSpec(object):
         assert len(self.types) == len(self.names), "len(%s) != len(%s)"%(self.types, self.names)
         #Header.msg support
         if (len(self.types)):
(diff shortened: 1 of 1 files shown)


dirk-thomas:
Also cherry-pick to hydro-devel.

I am not planning a new release for this patch since it is not significant enough to justify a rebuild.

@tfoote @wjwwood Please review.

wjwwood:
+1

tfoote:
+1

dirk-thomas:
Cherry-picked to _Hydro_ in f419e257ccf808d013a94091b9438f252de0748a

PR #69, opened 2014-10-26
fix OSError handling

dirk-thomas:
Regression of d1d3cc5514ae1d0f74f131598046fdb276035ebb

Similar to ros/ros_comm#516


tfoote:
+1

PR #70, opened 2014-10-30
Smarter tab-completion of rosbag in zsh

Enfenion:
The first commit, c94d7f7, simply cleans up existing roszsh by converting tabs to spaces and removing some out-commented code.
The second commit makes tab completion for `rosbag` more intelligent by completing with bag files for all commands except record which completes with the existing topics.

### DIFF
--- a/tools/rosbash/roszsh
+++ b/tools/rosbash/roszsh
@@ -383,6 +383,17 @@ function _roscomplete_rosbag {
                     ;;
             esac
             reply=(${=opts})
+	else
+	    case ${=${(s: :)words}[2]} in
+		record)
+		    opts=$(rostopic list 2>/dev/null)
+		    if [ -z "$opts" ]; then
+			reply="" #Prevent execution of next rule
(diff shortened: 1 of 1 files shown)


dirk-thomas:
We should choose one of the following two options for this PR:
- either limit the changes to functional patches only (which means no white space changes at all, in order to maintain the ability to easily back port future changes to other branches)
- or reformat the entire file regarding white spaces and indentation to be consistent, this should be a separate commit from any functional changes

Currently the PR mixes functional changes (9b50ec6) with white space changes (c94d7f7 and b199559) but the resulting file is still indented inconsistently (2 vs. 4 spaces).

Enfenion:
The whitespace issues should be fixed by 8089b30

dirk-thomas:
I manually merged the changes in 26366164c686de536fe3f1b43effb6a64951ef7f and 07df4a8433824dad86dcc50f803617bbf781fb62.

Issue #72, opened 2014-12-09
roscd behaviour without input is different between rosbash and roszsh


javanegmond:
Hi,

The rosbash version of roscd checks `CMAKE_PREFIX_PATH`, while the roszsh version does not. As a result, after source'ing (`setup.[bash|zsh]`, depending on the shell I use) the bash behaviour of `roscd` without parameters goes to `/path/to/workspace/devel`, while the zsh behavior is the following error:
`No ROS_WORKSPACE set.  Please set ROS_WORKSPACE to use roscd with no arguments.`

Is there a specific reason for this difference in behaviour?

dirk-thomas:
It is likely not intentional but has either been broken during recent updates to the zsh stuff or different people implementing the different shells weren't aware of it. Please feel free to provide a match to the `zsh` scripts to make it behave like `bash`.

javanegmond:
Alright, I believe my pull request fixes it. I checked and it also applies to hydro-devel. Should I make the changes for that and do another pull request?

I haven't checked rostcsh, it could be that changes in rosbash aren't reflected in there either.

dirk-thomas:
You don't have to create a separate PR for other distros. After it has been merged (and a potential soak time) it will be considered for being backported.

ROS workspace don't even have a setup.tcsh entry point anymore. I don't think it is used anymore anywhere.

dirk-thomas:
This has been cherry-picked to the _hydro-devel_ branch and released as 1.10.11.

Thank you!

PR #73, opened 2014-12-10
Changed behaviour of roscd in roszsh to match the roscd in rosbash

javanegmond:
My solution to #72 . Behaviour of the zsh roscd with no parameters now matches that of bash.


dirk-thomas:
Thanks, I have manually merged a modified version of your patch which has consistent intendation: 56eb5175e65109a44a4c1cf716571b3ac377af17

PR #74, opened 2014-12-31
Add argument to toggle caching of catkin_pkg.packages.find_packages

garaemon:
patch for https://github.com/ros/ros_comm/issues/541

### DIFF
--- a/core/roslib/src/roslib/packages.py
+++ b/core/roslib/src/roslib/packages.py
@@ -47,6 +47,10 @@ import string
 from subprocess import Popen, PIPE
 
 from catkin.find_in_workspaces import find_in_workspaces as catkin_find
+from catkin.find_in_workspaces import get_workspaces as catkin_get_workspaces
+from catkin.find_in_workspaces import get_source_paths as catkin_get_source_paths
+from catkin.find_in_workspaces import find_packages as catkin_find_packages
+
 import rospkg
 
 import roslib.manifest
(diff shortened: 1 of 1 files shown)


dirk-thomas:
Please consider to update the PR to use the existing argument `source_path_to_packages` of the `catkin_find` function (see https://github.com/ros/catkin/pull/704#issuecomment-68476827).

core/roslib/src/roslib/packages.py
```diff
@@ -509,3 +515,16 @@ def find_resource(pkg, resource_name, filter_fn=None, rospack=None):
...
+    Return packages which can be used in source_path_to_packages keyword
+    of catkin.find_in_workspaces.find_in_workspaces
+    """
+    source_path_to_packages = {}
+    workspaces = catkin_get_workspaces()
+    for workspace in workspaces:
+        source_spaces = catkin_get_source_paths(workspace)
+        for source_path in source_spaces:
+            source_path_to_packages[source_path] = catkin_find_packages(source_path)
+    return source_path_to_packages
```
dirk-thomas, line 530: The logic to iterate over the workspaces and finding packages in each of them is provided by catkin.find_in_workspaces and should be reused instead of being duplicated here.

dirk-thomas:
If you are still interested in this feature it would be highly appreciated if you could update the PR based on the comments above.

wkentaro:
@dirk-thomas 
#91 is another implementation for this. Please review it.

> The logic to iterate over the workspaces and finding packages in each of them is provided by catkin.find_in_workspaces and should be reused instead of being duplicated here.

Actually, I don't find the way to reuse function in catkin.find_in_workspaces so created another function.
https://github.com/ros/catkin/pull/757

PR #75, opened 2015-01-04
Fish shell support

aclough:
Hello, now that fish is in the standard Ubuntu repos I thought I'd re-submit this patch to enable fish support in ROS.  This adds all the standard shell functions, autocomplete for those, and autocomlete for a few of the more common or easier ros commands. 

I've tested rosfish itself but while I included changes to the various cmake files for the inclusion of rosfish I haven't been able to test those.  I would have, but building ros from the github checkout looked difficult and I didn't see any instructions online for it.

### DIFF
--- a/tools/rosbash/CMakeLists.txt
+++ b/tools/rosbash/CMakeLists.txt
@@ -3,7 +3,7 @@ project(rosbash)
 find_package(catkin)
 catkin_package()
 
-install(FILES rosbash rostcsh roszsh rosfish
+install(FILES rosbash rosfish rostcsh roszsh
   DESTINATION ${CATKIN_PACKAGE_SHARE_DESTINATION})
 install(PROGRAMS scripts/rosrun
   DESTINATION ${CATKIN_GLOBAL_BIN_DESTINATION})
--- a/tools/rosbash/CMakeLists.txt
+++ b/tools/rosbash/CMakeLists.txt
@@ -3,7 +3,7 @@ project(rosbash)
 find_package(catkin)
 catkin_package()
 
-install(FILES rosbash rostcsh roszsh
+install(FILES rosbash rostcsh roszsh rosfish
   DESTINATION ${CATKIN_PACKAGE_SHARE_DESTINATION})
 install(PROGRAMS scripts/rosrun
   DESTINATION ${CATKIN_GLOBAL_BIN_DESTINATION})
@@ -14,4 +14,4 @@ configure_file(${catkin_EXTRAS_DIR}/templates/script.bash.in
   ${CATKIN_DEVEL_PREFIX}/bin/rosrun
   @ONLY)
 
-catkin_add_env_hooks(15.rosbash SHELLS bash tcsh zsh DIRECTORY ${CMAKE_CURRENT_SOURCE_DIR}/env-hooks)
+catkin_add_env_hooks(15.rosbash SHELLS bash tcsh zsh fish DIRECTORY ${CMAKE_CURRENT_SOURCE_DIR}/env-hooks)
--- a/tools/rosbash/CMakeLists.txt
+++ b/tools/rosbash/CMakeLists.txt
@@ -14,4 +14,4 @@ configure_file(${catkin_EXTRAS_DIR}/templates/script.bash.in
   ${CATKIN_DEVEL_PREFIX}/bin/rosrun
   @ONLY)
 
-catkin_add_env_hooks(15.rosbash SHELLS bash tcsh zsh fish DIRECTORY ${CMAKE_CURRENT_SOURCE_DIR}/env-hooks)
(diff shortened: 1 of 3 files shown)


tools/rosbash/CMakeLists.txt
```diff
@@ -3,7 +3,7 @@ project(rosbash)
 find_package(catkin)
 catkin_package()
 
-install(FILES rosbash rostcsh roszsh
+install(FILES rosbash rostcsh roszsh rosfish
```
dirk-thomas: Please add the new entry in alphabetical order (same below).

tools/rosbash/rosfish
```diff
@@ -0,0 +1,357 @@
...
+function _rossed
+  if test (uname) = "Darwin" -o (uname) = "FreeBSD"
+      sed -E $argv
+  else
+      sed -r $argv
+  end
+end
+
+function _rosfind
+  if  test (uname) = "Darwin" -o (uanme) = "FreeBSD"
```
dirk-thomas: `uanme` looks wrong?

tools/rosbash/rosfish
```diff
@@ -0,0 +1,357 @@
...
+complete -c rosls -a '(_roscomplete_sub_dir)'
+complete -c rosls -l help --description "prints command info"
+complete -x -c rosmake -a '(_roscomplete_rosmake)'
+complete -x -c rosclean -a 'check purge'
+complete -x -c rosrun -a '(_roscomplete_rosrun)'
+complete -x -c roslaunch -a '(_roscomplete_launch)'
+complete -x -c rosed -a '(_roscomplete_file)'
+complete -c rosed -l help --description "prints command info"
+complete -x -c roscp -a '(_roscomplete_file)'et -x TEST works
+complete -c roscp -l help --description "prints command info"
```
dirk-thomas, line 357: Please use a consistent indentation across the whole file - either 2 or 4 spaces.

dirk-thomas:
A general question about _fish_: how much different are these scripts to the existing ones (bash / zsh)? Are there any general incompatibilities or does the new file add fish-specific features not available in bash / zsh? Could one of the existing ones be updated to also work in fish?

Please also build the changed packages from source and actually run the code afterwards to ensure that it works as expected. The general from-source instructions should be a good starting point: http://wiki.ros.org/indigo/Installation/Source You might also want to look at the tutorials of the buildsystem: http://wiki.ros.org/catkin/Tutorials

tools/rosbash/rosfish
aclough: That's just how fish syntax works.

aclough:
The syntax for fish is actually different from standard shell syntax so I can't just run bash or zsh shell scripts I'm afraid.  Using (uname) instead of `uname` for example. 

I looked at the instructions you linked when thinking about how to try building ros from source with my modifications but gave up when the instructions didn't seem to cover having local source code rather than downloading the source.  Re-reading it, though, I actually think it should be pretty easy to figure out the right place to depart from the instructions so I'll go ahead and do that.  I've been reading up on catkin recently since our company is just about to move from Fuerte to Indigo.

tools/rosbash/rosfish
tfoote: `uanme` not `uname`?

aclough: Ah, right, that's a problem.

aclough:
Tested it out with a catkin_make_isolated and everything shows up in install_isolated/share/rosbash just as it should.  Are there other aspects of the install process I should be testing here?

tools/rosbash/CMakeLists.txt
```diff
@@ -14,4 +14,4 @@ configure_file(${catkin_EXTRAS_DIR}/templates/script.bash.in
   ${CATKIN_DEVEL_PREFIX}/bin/rosrun
   @ONLY)
 
-catkin_add_env_hooks(15.rosbash SHELLS bash tcsh zsh DIRECTORY ${CMAKE_CURRENT_SOURCE_DIR}/env-hooks)
+catkin_add_env_hooks(15.rosbash SHELLS bash tcsh zsh fish DIRECTORY ${CMAKE_CURRENT_SOURCE_DIR}/env-hooks)
```
dirk-thomas: This should also be in alphabetical order.

dirk-thomas:
How do you plan to use the fish-specific scripts? Should `catkin` also create a top-level `setup.fish`?

aclough:
What I do for myself is to just set the necessary environmental variables directly, it was pretty straightforward.  I think it would be possible to create a setup.fish that works like setup.bash (though it would require a few sed invocations to get the formatting right) but I'd rather finish the tab completions first.

dirk-thomas:
Sounds good for now. Can you please squash your commit (which keeps it easy to cherry-pick it between branches).

PR #76, opened 2015-01-22
rosclean: Use "du -sk * 1024" on Linux for getting disk usage since "du ...

aballier:
...-b" is not supported by busybox du, while "du -k" is supported by both coreutils and busybox.

### DIFF
--- a/tools/rosclean/src/rosclean/__init__.py
+++ b/tools/rosclean/src/rosclean/__init__.py
@@ -121,13 +121,13 @@ def get_disk_usage(d):
     """
     Get disk usage in bytes for directory
     :param d: directory path, ``str``
-    :returns: disk usage in bytes (du -b) or (du -A) * 1024, ``int``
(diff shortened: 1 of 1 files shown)


dirk-thomas:
`du -sk` multiplied with `1024` returns a very different value than `du -sb` for me?

aballier:
Hmm yes, indeed, that is because of --apparent-size enabled with -b:

```
    --apparent-size
              print apparent sizes, rather than disk usage; although the apparent size is usually smaller, it may be larger due to holes in ('sparse') files, internal fragmentation, indirect blocks, and the like
```

Is that really what is wanted ? The description of the function talks about "disk usage" but then again, du -A on FreeBSD is also about apparent size, so I'm a bit lost.
I don't see any option to have apparent size on busybox's du, so if apparent size is really what is wanted it means it cannot work with busybox's du, which is sad because this is usually the standard on embedded systems.

at-wat:
@aballier FYI, #185 added support for busybox du

PR #77, opened 2015-01-27
Adding support for fish shell

aclough:
This adds a rosfish alongside rosbash and the others to provide shell
functions such as roscd and appropriate tab completions.  Due to
differing syntax fish users will have to set environemnt variables
manually that are set by setup.sh for bash and zsh but once that is done
they should be able to use all the same facilities as other supported
shells.

### DIFF
--- a/tools/rosbash/CMakeLists.txt
+++ b/tools/rosbash/CMakeLists.txt
@@ -3,7 +3,7 @@ project(rosbash)
 find_package(catkin)
 catkin_package()
 
-install(FILES rosbash rostcsh roszsh
+install(FILES rosbash rosfish rostcsh roszsh
   DESTINATION ${CATKIN_PACKAGE_SHARE_DESTINATION})
 install(PROGRAMS scripts/rosrun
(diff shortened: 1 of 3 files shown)


dirk-thomas:
Replaces #75.

Next time you can simply force push the squashed commit to the same branch the PR is made from. It will update automatically.

ruffsl:
Can this be cherry-pick'ed into indigo-devel? As a fish shell user, it'd be nice to see this in the current LTS.

dirk-thomas:
I  cherry-picked the commit (2901e4a3401e62b48c8dc1282947cdf401deb529) but it will likely take quite a while until it will be available from Debian since this repo was recently released into Indigo and no important fixes are pending atm. This change alone doesn't justify a new release (which will imply a rebuild of all downstream packages on the farm and a redownload of all downstream packages by users). Also the next LTS is already close (May 2016).

ruffsl:
Understood, thanks dirk-thomas and aclough.

PR #80, opened 2015-04-03
remove usage of CATKIN_TEST_RESULTS_DIR environment variable

dirk-thomas:
Relates to ros/catkin#728.


core/roslib/src/roslib/rosenv.py
```diff
@@ -60,7 +60,6 @@
 ## directory in which log files are written
 ROS_LOG_DIR      ="ROS_LOG_DIR"
 ## directory in which test result files are written
```
wjwwood: Should the comment also go?

dirk-thomas: Indeed...

PR #87, opened 2015-08-05
Fix ros/catkin#694

gerkey:
@tho-, @kartikmohta, @mikey0000 for feedback; please let us know whether this fix works for you on your Ubuntu >=14.10 systems.

**Two options for testing this fix:**
1. Temporarily overwrite your `/opt/ros/jade/share/ros/core/rosbuild/public.cmake` with the modified version from this branch; OR
2. check out this branch of `ros` and set your `ROS_ROOT` to point to `ros/core/rosbuild`.

Then try building a rosbuild package.

---

**Nature of the change:**

Modify rosbuild to remove the leading `:` (colon) prefix from any absolute library path that was returned by rospack.  Where they are present, these prefixes were inserted by catkin when generating `.pc` files; these files are processed by pkg-config, which is invoked by rospack to get build flags.  It would be nice to remove the `-l:` prefix from the `.pc` files, but they need to stay there to ensure consistent ordering in the returned strings from multiple versions of pkg-config (0.26 and 0.28 have totally different logic, leading to different results).

So instead, we leave the `-l:` prefixes in the `.pc` files, which gives us the right order of libs in rosbuild.  Then we strip the prefixes before passing the libs, still in the right order, to `target_link_libraries()`.  While newer versions of ld won't accept `-l:/path/to/mylib.so`, both old and new versions of ld will accept `-l/path/to/mylib.so`, so doing this strip is safe.

There is one noticeable side effect of this change, which is how libraries in system directories are handled.  Previously, we were passing things like `:/usr/lib/x86_64-linux-gnu/libboost_signals.so` to `target_link_libraries()`.  CMake passed that straight through to ld as `-l:/usr/lib/x86_64-linux-gnu/libboost_signals.so`.  Now we're passing `/usr/lib/x86_64-linux-gnu/libboost_signals.so` to `target_link_libraries()`.  CMake doesn't pass that straight through, but rather recognizes the library as being in a system location and instead strips it down to `-lboost_signals` before passing to ld.   This mechanism and the rationale behind it are explained in the [CMake 2.6 release notes](http://www.vtk.org/Wiki/CMake_2.6_Notes#Linking_to_System_Libraries).  After reading that rationale, I believe that the new behavior is an improvement for us over the old behavior.  In any case, it should also be safe.

**Edit:** the above special handling of libraries in system directories does not affect libraries in other directories, which is important for supporting catkin workspace overlays.  E.g., if we put `-l:/opt/ros/jade/lib/libfoo.so` in a `.pc` file, then rosbuild will see `:/opt/ros/jade/lib/libfoo.so`, strip the leading `:`, and pass `/opt/ros/jade/lib/libfoo.so` to `target_link_libraries()`, which will in turn pass the full path `/opt/ros/jade/lib/libfoo.so` to ld.  @dirk-thomas and I verified this behavior manually (I'd love to provide an automated test, but it's logistically very difficult to write tests for rosbuild).

### DIFF
--- a/core/rosbuild/core/rosbuild/public.cmake
+++ b/core/rosbuild/core/rosbuild/public.cmake
@@ -258,6 +258,25 @@ macro(rosbuild_init)
       list(REVERSE ${_prefix}_LIBRARIES)
     endif()
 
+    # Strip the leading ':' from any library that was given to us as
+    # ':/path/to/lib'.  Those kinds of libraries generally come from
+    # catkin-generated .pc files that are parsed by rospack via pkg-config.
+    # Newer versions of ld (and gold) don't accept the '-l:/path/to/lib'
+    # syntax.  They've long accepted the '/path/to/lib' syntax, so it
+    # should be safe to just strip the ':' here.  We could remove it at
+    # the source, from the .pc file, but the lib argument needs to start
+    # with '-l' to ensure proper ordering of libs in the return from
+    # pkg-config.
+    # https://github.com/ros/catkin/issues/694
+    if(${_prefix}_LIBRARIES)
+      set(_${_prefix}_LIBRARIES ${${_prefix}_LIBRARIES})
+      set(${_prefix}_LIBRARIES "")
+      foreach(_lib ${_${_prefix}_LIBRARIES})
+        string(REGEX REPLACE "^:/" "/" _lib_nocolon ${_lib})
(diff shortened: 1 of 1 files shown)


mikey0000:
I'll try it this Friday

kartikmohta:
This fixed the issue for me.

PR #89, opened 2015-09-17
Removing invalid characters from XML unit test results

pjreed:
Previously, rosunit was dumped the unfiltered stdout/stderr from tests into its XML result files.  They often contained ASCII control codes (such as for colorizing output) which are invalid in XML, and so validating XML parsers that tried to ingest the test results would fail to the parse the files.

I modified the XML generation code to filter out nonprintable characters from the stdout/stderr, and while I was at it, I also modified it to use lxml to generate XML rather than simply appending strings, so other special characters will be automatically escaped.

### DIFF
--- a/tools/rosunit/src/rosunit/junitxml.py
+++ b/tools/rosunit/src/rosunit/junitxml.py
@@ -49,11 +49,21 @@ import string
 import codecs
 import re
 
+import xml.etree.ElementTree as ET
 from xml.dom.minidom import parse, parseString
 from xml.dom import Node as DomNode
 
 from functools import reduce
 import rospkg
+
+invalid_chars = re.compile(ur'[^\x09\x0A\x0D\x20-\x7E\x85\xA0-\xFF\u0100-\uD7FF\uE000-\uFDCF\uFDE0-\uFFFD]')
+def invalid_char_replacer(m):
+    return "&#x"+('%04X' % ord(m.group(0)))+";"
+def filter_nonprintable_text(text):
+    return re.sub(invalid_chars, invalid_char_replacer, text)
+
+def cdata(cdata_text):
+    return '<![CDATA[\n{}\n]]>'.format(cdata_text)
 
 class TestInfo(object):
     """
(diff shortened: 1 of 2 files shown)


dirk-thomas:
It sounds like a good idea to ensure to produce valid xml output but I have a few comments regarding this patch:
- It changes the arguments / semantic of existing stable API. This can't be released into existing distributions and even for the next upcoming one I am not sure if the change is worth the effort to update existing code using this API.
- Since `lxml` is not part of a default Python installation I am a bit hesitant to introduce it as a dependency.
- All problematic characters are simply removed. Wouldn't it be better to keep them but escaped instead?

pjreed:
- What else uses this API that would need to change?  Or rather, what in the API changed that would cause problems?  I haven't encountered any issues when using this in my own testing, but I can modify it to be compatible with the previous behavior if there's an issue.
- I had lxml already installed on my OS and didn't realize it wasn't part of the default Python install.  I can easily switch this to Python's default etree implementation, which should work fine.
- I'm not aware of an official way to escape ASCII control characters in valid XML.  Do you have a recommended way to do that?

dirk-thomas:
I don't have a specific code breaking because of the change. But arbitrary code might exist in ROS packages which uses the existing API. If the arguments or their semantic changes that existing code will break.

Yes, using `xml.etree` would be better.

I thought about a `CDATA` section but the escaping necessary inside might still be some effort (https://en.wikipedia.org/wiki/CDATA).

dirk-thomas:
I marked this with the milestone "untargeted" until there is any update.

pjreed:
After some more work:
- The existing rosunit API is preserved
- It now uses the built-in Python ElementTree implementation
- Non-printable characters are properly replaced as Unicode control codes in a CDATA section

dirk-thomas:
The diff shows many unrelated changed likely due to a wrong merge. Please rebase the patch against the current state of the devel branch.

pjreed:
I've rebased against jade-devel and cleaned up the commit history.

dirk-thomas:
Thanks for the patch and updating it.

tools/rosunit/src/rosunit/junitxml.py
```diff
@@ -75,25 +85,42 @@ class TestError(TestInfo):
...
-        try:
-            data = unicode(data)
-        except NameError:
-            pass
-        return data
+        """
+        @return XML tag representing the object, with non-XML text filtered out
+        @rtype: str
+        """
+        return ET.tostring(self.xml_element, encoding='utf-8', method='xml')
```
evenator, line 113: Shouldn't this line be

``` python
        return ET.tostring(self.xml_element(), encoding='utf-8', method='xml')
```

tools/rosunit/src/rosunit/junitxml.py
```diff
@@ -75,25 +85,42 @@ class TestError(TestInfo):
...
```
evenator, line 92: Shouldn't this line be

``` python
        return ET.tostring(self.xml_element(), encoding='utf-8', method='xml')
```

tools/rosunit/src/rosunit/junitxml.py
```diff
@@ -169,15 +196,26 @@ def add_error(self, error):
...
+        @rtype: xml.etree.ElementTree.Element
+        """
+        testcase = ET.Element('testcase')
+        testcase.set('classname', self.classname)
+        testcase.set('name', self.name)
+        testcase.set('time', str(self.time))
+        for f in self.failures:
+            testcase.append(f.xml_element())
+        for e in self.errors:
+            testcase.append(e.xml())
```
evenator, line 217: Shouldn't this line be

``` python
            testcase.append(e.xml_element())
```

dirk-thomas:
It seems like this change might be responsible for the following regression: http://build.ros.org/job/Kpr__ros_comm__ubuntu_xenial_amd64/1/consoleFull search for `IndexError: list index out of range`

PR #92, opened 2015-10-06
rosbash: Fix zsh rosservice completion.

de-vri-es:
I had to change this for rosservice completion to work correctly. `${words[-1]}` is the message value (possibly empty string) at this point and not the service name. I changed it to `${words[3]}`. `${words[-2]}` should also work, but since the subcommand is already hardcoded as `${words[2]}` and there are no optional options I saw no harm in hardcoding index 3 as well.

Also I am very curious why the words array is being accessed sometimes as `${=${(s: :)words[x]}}`. It seems like a very elaborate NOP to me. Why do that over `${words[x]}`?

Alternatively `${=${(s: :)words}[x]}` is not a NOP (though the second split might be) but seems plain wrong to me. We should not be doing word splitting at all, the shell has already done that while taking into account quotes and escape characters and what not. Flattening the array into a string and then splitting on spaces is simply wrong, not to mention more complex and hard to read.

/edit:

The complection that was not working was:
`rosservice call /some/service [tab]`

It should give the layout of the request body so you can fill in details. However, it just wrote errors to stderr. After the PR it should fill in the request body.

### DIFF
--- a/tools/rosbash/roszsh
+++ b/tools/rosbash/roszsh
@@ -565,7 +565,7 @@ function _roscomplete_rosservice {
         opts="list call type find uri"
         reply=(${=opts})
     elif [[ ${CURRENT} == 3 ]]; then
-        case ${=${(s: :)words}[2]} in
+        case ${words[2]} in
             uri|list|type|call)
                 opts=`rosservice list 2> /dev/null`
                 IFS=$'\n'
@@ -573,16 +573,15 @@ function _roscomplete_rosservice {
                 unset IFS
                 ;;
             find)
-                opts=`_srv_opts ${=${(s: :)words[-1]}}`
+                opts=`_srv_opts ${words[3]}`
                 reply=(${=opts})
                 ;;
         esac
     elif [[ ${CURRENT} == 4 ]]; then
-        case ${=${(s: :)words}[2]} in
+        case ${words[2]} in
             call)
-                type=`rosservice type ${=${(s: :)words[-1]}}`
-                opts=`rosmsg-proto srv 2> /dev/null -s ${type}`
-                reply=(${=opts})
+                type="$(rosservice type ${words[3]} 2>/dev/null)"
+                reply=("$(rosmsg-proto srv -s ${type} 2> /dev/null)")
                 ;;
         esac
     fi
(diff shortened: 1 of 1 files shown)


wjwwood:
Near as I can tell this style of dereferencing the "words" was introduced here: https://github.com/ros/ros/commit/333f347749d6e4c4e3031f3ab764a1ecc24b56f4

After that everyone seems to just have mimicked the style. I'm a little wearing changing it, since I'm not sure why it was done in the first place.

de-vri-es:
I see. Thanks for the pointer.

@kwc: Could you elaborate why you decided to split `$words` like that? Is there something I overlooked?

tools/rosbash/roszsh
```diff
@@ -740,7 +739,7 @@ compctl -K "_roscomplete_rosbag" + -g "*.bag *(/)" "rosbag"
 compctl -K "_roscomplete_rosnode" "rosnode"
 compctl -K "_roscomplete_rosparam" "rosparam"
 compctl -x 'p[0,2]' -K "_roscomplete_rostopic" - 'n[1,/] p[3]' -K "_roscomplete_rostopic" - 'p[3]' -S ' ' -K "_roscomplete_rostopic" - 'p[4]' -Q -K "_roscomplete_rostopic" -- "rostopic"
-compctl -K "_roscomplete_rosservice" "rosservice"
+compctl -Q -K "_roscomplete_rosservice" "rosservice"
```
wjwwood, line 742: Why did you add `-Q`?

wjwwood:
@de-vri-es, @kwc may not have time to answer, but we can try to reason about why that be necessary. I'm not opposed to changing it, but it would be great if we could come up with a reason as to why someone might do that. It also looks like @kwc may have just merged someone else's patch, so it might not be something he did originally.

de-vri-es:
(should have commented in the line notes)

tools/rosbash/roszsh
de-vri-es, line 742: `-Q` prevents zsh from escaping special characters like quotes and spaces in the completion candidates. Since rosmsg-proto already gives output suitable for parsing by the shell, escaping it just makes it illegible

de-vri-es:
@wjwwood: I see. I can't really think of a reason though. What will happen is that arguments containing spaces on the command line (escaped with a backslash or quotes) will be turned into separate array elements. In the original `$words` array, the argument will be just one array element, even if it contains spaces.

However, the behaviour of the original `$word` array without splitting seems correct to me. After all, that array truly represents the separate command line arguments.

I can test locally with just `${word[i]}` and see if anything breaks. I doubt I will trigger every code path during my normal usage though.

wjwwood:
@de-vri-es I think it would be ok to take the pr as proposed after we test it locally, but I'll let @dirk-thomas voice an opinion before doing that.

In the mean time, could you give an example of what completion was not working before and that works after this patch so that I can try it on my machine?

de-vri-es:
@wjwwood: Ah yes, should have done that in the PR.

The complection that was not working was:
`rosservice call /some/service [tab]`

It should give the layout of the request body so you can fill in details. However, it just wrote errors to stderr. After the PR it should fill in the request body.

wjwwood:
I just tested this against indigo on my mac and I can confirm the before and after behavior.

dirk-thomas:
Same for me with `zsh` on Ubuntu.

Thank you for the patch.

de-vri-es:
@wjwwood @dirk-thomas: Thanks for the quick tests and merge :)

PR #93, opened 2015-10-06
robash: Fix zsh rosservice completion.

de-vri-es:
Same as #92 but for indigo-devel instead of jade-devel.


dirk-thomas:
Thank you, but I will backport the commits from jade-devel manually to other branches before patch releases. Therefore I will close this.

de-vri-es:
Righty, no problem. Thanks again.

PR #94, opened 2015-10-08
rosrun: Allow spaces in command names and search paths.

### LINKED ISSUE #85: rosrun completion incorrect for files with spaces
When tab-completing a filename with spaces with rosrun, the space in the filename is not escaped properly.

For example:

```
touch 'foo bar'
rosrun roscpp_tutorials talker foo<TAB>
```

Completes to:

```
rosrun roscpp_tutorials talker foo bar
```

Instead, it should complete with the space escaped:

```
 rosrun roscpp_tutorials talker foo\ bar
```

(obviously the talker node doesn't take files as arguments, but the example serves to demonstrate how tab-completion is incorrect).

de-vri-es:
This PR allows command names and search paths for `rosrun` to contain spaces. It mainly adds quotes around a bunch of variables and sets `IFS=$'\n'` on two places to split only on newlines.

In the case of `$catkin_package_libexec_dr`, I turned it into an array (also by splitting the output with `IFS=$'\n'`, and renamed it to be plural) so the individual elements can contain spaces when passing it to `find`.

I tested with a test packages containing one script with a space in the name:
`rosrun test_package "test script"`

Without the PR, `rosrun` errors out on a `find` command. WIth the PR it works as expected. The same goes for workspace paths containing spaces, previously they would have caused `find` to error, but with the PR it works.

Does not fix #85 but it makes sense to do this first ;)

### DIFF
--- a/tools/rosbash/scripts/rosrun
+++ b/tools/rosbash/scripts/rosrun
@@ -16,13 +16,13 @@ function debug() {
 
 while [ $args == 1 ]
 do
-  case $1 in
+  case "$1" in
     "--help" | "-h")
       usage
       exit 0
       ;;
     "--prefix" | "-p")
-      rosrun_prefix=$2
+      rosrun_prefix="$2"
       shift
       shift
       ;;
@@ -48,12 +48,14 @@ case $2 in
 esac
 
 if [[ -n $CMAKE_PREFIX_PATH ]]; then
-  catkin_package_libexec_dir=`catkin_find --without-underlays --libexec --share $1 2> /dev/null`
-  debug "Looking in catkin libexec dir: $catkin_package_libexec_dir"
+  IFS=$'\n'
+  catkin_package_libexec_dirs=(`catkin_find --without-underlays --libexec --share "$1" 2> /dev/null`)
+  unset IFS
+  debug "Looking in catkin libexec dirs: $catkin_package_libexec_dirs"
 fi
-pkgdir=`rospack find $1`
+pkgdir=`rospack find "$1"`
 debug "Looking in rospack dir: $pkgdir"
-if [[ -z $catkin_package_libexec_dir && -z $pkgdir ]]; then
+if [[ ${#catkin_package_libexec_dirs[@]} -eq 0 && -z $pkgdir ]]; then
   exit 2
 fi
 if [[ ! $2 == */* ]]; then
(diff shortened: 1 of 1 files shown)


tools/rosbash/scripts/rosrun
```diff
@@ -16,13 +16,13 @@ function debug() {
 
 while [ $args == 1 ]
 do
-  case $1 in
+  case "$1" in
```
dirk-thomas, line 19: The first argument is the package name which can by definition not contain spaces.

tools/rosbash/scripts/rosrun
```diff
@@ -48,12 +48,14 @@ case $2 in
 esac
 
 if [[ -n $CMAKE_PREFIX_PATH ]]; then
-  catkin_package_libexec_dir=`catkin_find --without-underlays --libexec --share $1 2> /dev/null`
-  debug "Looking in catkin libexec dir: $catkin_package_libexec_dir"
+  IFS=$'\n'
```
dirk-thomas, line 51: What is the `$` for? Shouldn't it be just `IFS='\n'`.

Same below.

tools/rosbash/scripts/rosrun
de-vri-es, line 19: Fair enough, but it can't hurt either.

tools/rosbash/scripts/rosrun
de-vri-es, line 51: `$'\n'` is the easiest way to get a literal newline in bash. You can try it yourself:

```
$ foo='\n'
$ bar="\n"
$ echo $foo$bar
```

prints: `\n\n`

dirk-thomas, line 51: Thank for clarifying. Its good as it is then. I usually use:

```
IFS="
"
```

tools/rosbash/scripts/rosrun
de-vri-es, line 19: Actually, at this point the arguments may also be options (`--help` or `--prefix`). Not that those contain spaces, but the user could still type them. Even on invalid input the program should behave nice and display the proper error message, I think.

Though as it turns out bash doesn't do word splitting on the argument to `case` so it still isn't necessary here. But it also can't hurt either.

tools/rosbash/scripts/rosrun
```diff
@@ -66,15 +68,18 @@ if [[ ! $2 == */* ]]; then
     _perm="/111"
   fi
   debug "Searching for $2 with permissions $_perm"
-  exepathlist=(`find -L $catkin_package_libexec_dir $pkgdir -name $2 -type f  -perm $_perm ! -regex ".*$pkgdir\/build\/.*" | uniq`)
+  exepathlist="`find -L "${catkin_package_libexec_dirs[@]}" "$pkgdir" -name "$2" -type f  -perm "$_perm" ! -regex ".*$pkgdir\/build\/.*" | uniq`"
```
dirk-thomas, line 71: Since `${catkin_package_libexec_dirs[@]}` might result in multiple paths `find -L "/path/foo /path/bar" ...` will fail.

dirk-thomas, line 71: Running the `find` command manually fails. Within the `rosrun` script is seems to deal with the fact that two paths are quotes as one...

de-vri-es, line 71: `find -L` means follow symbolic links. It doesn't take any arguments. Any non-option arguments before predicates are search paths for find.

Additionaly `"${array[@]}"` (quotes are important)  expands to one word (argument) per element. The previous version also resulted in multiple arguments, but split on any whitespace character instead of one argument per array element. So in the case of paths containing spaces the old version actually resulted (incorrectly) in more arguments.

de-vri-es, line 71: Ah, I think there is a slight misconception here how arrays are expanded.

Suppose you have the array `array=("aap noot mies" "wim zus jet")`.
Then `"${array[@]}"` expands to `"aap noot mies" "wim zus jet"` (one word per array element) and not to `"aap noot mies wim zus jet"`.

That's the entire reason I switched `$catkin_package_libexec_dirs` to an array.

dirk-thomas, line 71: Thank you for clarifying it. It took me a bit how the quoting works (https://gist.github.com/dirk-thomas/b7c9c77a996518c2fe5e).

dirk-thomas:
Thank you for the patch.

dirk-thomas:
Also cherry-picked to the _indigo-devel_ branch: 6cef8176f6fd9df476aaa6341effe20591995ec5

de-vri-es:
Thanks for the merge :)

PR #99, opened 2015-12-21
[rosbash] 	Add roscat to display file contents as stdout

wkentaro:
In most of my cases, I don't want to edit file but just view.
That's why I'm adding this function, the candidates of names:
- rosless
- rosview

### DIFF
--- a/tools/rosbash/rosfish
+++ b/tools/rosbash/rosfish
@@ -1,3 +1,4 @@
+#alias roscat 'echo "unimplemented"'
 
 function _rossed
     if test (uname) = "Darwin" -o (uname) = "FreeBSD"
--- a/tools/rosbash/rosbash
+++ b/tools/rosbash/rosbash
@@ -282,6 +282,23 @@ function roscp {
     cp ${arg} ${3}
 }
 
+function roscat {
+    local arg
+    if [[ $1 = "--help" ]] || [[ $# -ne 2 ]]; then
+       echo -e "usage: roscat [package] [file]\n\nDisplay a file content within a package."
+       [[ $1 = "--help" ]] && return 0 || return 1
+    fi
+    _roscmd ${1} ${2}
+    [ $? -eq 1 ] && return 1
+    if [[ -n ${arg} ]]; then
+        if [[ -z $CATTER ]]; then
+            cat ${arg}
+        else
+            $CATTER ${arg}
+        fi
+    fi
+}
+
 function rosawesome {
    alias megamaid='rosbag record'
    alias suck2blow='rosbag play'
@@ -949,6 +966,7 @@ complete -F "_roscomplete_rosclean" "rosclean"
 complete -F "_roscomplete_exe" "rosrun"
 complete -F "_roscomplete_file" "rosed"
 complete -F "_roscomplete_file" "roscp"
+complete -F "_roscomplete_file" "roscat"
 complete -F "_roscomplete_launch" -o filenames "roslaunch"
 complete -F "_roscomplete_test" -o filenames "rostest"
 complete -F "_roscomplete_rospack" "rospack"
--- a/tools/rosbash/rostcsh
+++ b/tools/rosbash/rostcsh
@@ -4,6 +4,7 @@ alias rosls 'ls `rospack find \!*`'
 #alias rosd 'echo "unimplemented"'
 #alias rospd 'echo "unimplemented"'
 #alias rosed 'echo "unimplemented"'
+#alias roscat 'echo "unimplemented"'
 #alias roscmd 'echo "unimplemented"'
 #alias roscp 'echo "unimplemented"'
 
--- a/tools/rosbash/roszsh
+++ b/tools/rosbash/roszsh
@@ -263,6 +263,21 @@ function roscp {
     cp ${arg} ${3}
 }
 
+function roscat {
+    local arg
+    if [[ $1 = "--help" ]] | [[ $# -ne 2 ]]; then
+        echo -e "usage: roscat [package] [file]\n\nDisplay a file content within a package."
+        [[ $1 = "--help" ]] && return 0 || return 1
+    fi
+    _roscmd ${1} ${2}
+    [ $? -eq 1 ] && return 1
+    if [[ -z $CATTER ]]; then
+        cat ${arg}
+    else
+        $CATTER ${arg}
+    fi
+}
+
 function _roscomplete {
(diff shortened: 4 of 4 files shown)


tfoote:
Starting with the simpler problem of just displaying the file might it be cleaner. Just implement `roscat` and then if you want to view it paginated you could do `roscat foo bar | less`  or `rosless` could be implemented extending that capability.

wkentaro:
That would be better :+1:

wkentaro:
I updated the commit.

tfoote:
That looks a lot cleaner. @dirk-thomas it looks good to me but I haven't had a chance to test it.

dirk-thomas:
I tried `rosview rospy_tutorials listener` in bash which didn't output anything. I also tried zsh with the command `rosview rospy_tutorials listener` which printed `That file does not exist in that package.`. So I don't think this is ready to be merged.

I also think `rosview ...` which simply does `roscat ... | less` is not necessary. We don't have to reinvent every command line tool to save the user to type a handful of letters on the console.

wkentaro:
I fixed the code for rosbash.
Actually I cannot reproduce the problem on zsh.

Also should I remove roscat?

dirk-thomas:
I would suggest to remove `rosview`. But lets see that others think, @ros/ros_team ?

tfoote:
I think roscat is valuable, since there's no quick way to view files.  But I agree that rosview seems redundant with roscat ... | less and is probably worth avoiding the extra complexity for future maintenance.

wjwwood:
I'd give +1 to both commands. While I understand that `rosview` can easily be represented with an alias or simply typing `roscat <args> | less` and that there is some maintenance cost, I still think taking the contribution is fine.

dirk-thomas:
I have tried `roscat` again and it works well when passing two argument in bash ans zsh (I haven't tested fish). But when calling `roscat` without arguments it prints `Couldn't find package []` and also has a return code of zero. I think in that case it should print the usage as well as return 1 (since --help was not explicitly requested). Looking at the patch it seems to aim for that behavior but the condition is wrong.

Regarding `rosless`: please remove it from the PR since only one out of three reviews seems to be in favor.

Thank you for iterating on this!

wkentaro:
Thanks, for the comments!
I have updated the commit. (removed rosview, added condition to show help when there're few arguments)

tools/rosbash/rosfish
```diff
@@ -279,6 +279,26 @@ function roscp
     cp $arg $argv[3]
 end
 
+function roscat
+    set -l arg
+    if test (count $argv) = 0
```
dirk-thomas: This condition hasn't been updated. Have you tried this case as it is? Does this work with fish?

wkentaro: Oops, I overlooked fish, will soon fix it.

2016年3月3日木曜日、Dirk Thomasnotifications@github.comさんは書きました:

> In tools/rosbash/rosfish
> https://github.com/ros/ros/pull/99#discussion_r54814099:
> 
> > @@ -279,6 +279,26 @@ function roscp
> >      cp $arg $argv[3]
> >  end
> > 
> > +function roscat
> > -    set -l arg
> > -    if test (count $argv) = 0
> 
> This condition hasn't been updated. Have you tried this case as it is?
> Does this work with fish?
> 
> —
> Reply to this email directly or view it on GitHub
> https://github.com/ros/ros/pull/99/files#r54814099.

## 

和田 健太郎 / Kentaro Wada
http://wkentaro.com

wkentaro:
I updated the commit, but cannot try to use on fish because I don't know how to use ros with fish. (there's no setup.fish, is it?)

dirk-thomas:
You can just source the file you modified directly in a fish shell and should be able to use the commands defined in there.

wkentaro:
There is the below error even if on jade-devel branch.

```
$ rosed rospy_tutorials listener
test: unexpected argument at index 2: '/opt/ros/indigo'
test: Expected a combining operator like '-a' at index 2
That file does not exist in that package.
```

dirk-thomas:
Can you please look into a fix for this?

This should be backported to Indigo once it is merged.

wkentaro:
hmm, I can't find how to fix it.
I'm not familiar with fish shell..

wkentaro:
I removed the code for fish.
I can't believe the rosfish is working:

```
wkentaro@anaconda ~/r/i/s/r/r/t/rosbash> exec fish --login
wkentaro@anaconda ~/r/i/s/r/r/t/rosbash> . rosfish
complete: Too many arguments
/home/wkentaro/ros/indigo/src/ros/ros/tools/rosbash/rosfish (line 357): complete -x -c roscp -a '(_roscomplete_file)'et -x TEST works
                                                                        ^
in . (source) call of file “/home/wkentaro/ros/indigo/src/ros/ros/tools/rosbash/rosfish”,
    called on standard input,


       complete - edit command specific tab-completions

   Synopsis
       complete (-c|--command|-p|--path) COMMAND [(-s|--short-option)
       SHORT_OPTION] [(-l|--long-option|-o|--old-option) LONG_OPTION
       [(-a||--arguments) OPTION_ARGUMENTS] [(-d|--description) DESCRIPTION]

complete: Type “help complete” for related documentation
```

dirk-thomas:
The return code still needs some work. When running e.g. `roscat roscpp_tutorials` and the usage is printed it should not have a return code of `0`. Same for calling it with a not existing file as well calling `roscat` without arguments.

But calling `roscat --help` should return `0`.

wkentaro:
Thanks, I fixed it.

dirk-thomas:
`roscat roscpp_tutorials foo` still returns 0.

wkentaro:
I fixed it. Please review again.

On Thu, Mar 10, 2016 at 2:30 AM, Dirk Thomas notifications@github.com
wrote:

> roscat roscpp_tutorials foo still returns 0.
> 
> —
> Reply to this email directly or view it on GitHub
> https://github.com/ros/ros/pull/99#issuecomment-194414256.

## 

和田 健太郎 / Kentaro Wada
http://wkentaro.com

dirk-thomas:
That looks good now. Can you please squash the commits before the merge. Thank you!

wkentaro:
Squashed.

dirk-thomas:
Cherry-picked to indigo-devel: 696efa7fdd0e370be7513dd6baeac06a577fe116

wkentaro:
Thanks.

PR #102, opened 2016-02-01
Expose an api which can map multiple export values to one package name

stonier:
The map based api has a problem - it loses export values when there are more than one per package name (simply overwrites any that were previously stored with the same package name key). This doesn't match the behaviour of rospack from the command line for which it is possible to list up multiple export values for each package name.

I've marked the map based api as deprecated - it really shouldn't be in there if it gives erroneous results, but some folk may be using it.


dirk-thomas:
I have created #106 to replace this. I applied some minor style changes

Thank you for providing this patch!

PR #103, opened 2016-02-01
Expose an api which can map multiple export values to one package name.

stonier:
The map based api has a problem - it loses export values when there are more than one per package name (simply overwrites any that were previously stored with the same package name key). This doesn't match the behaviour of rospack from the command line for which it is possible to list up multiple export values for each package name.

I've marked the map based api as deprecated - it really shouldn't be in there if it gives erroneous results, but some folk may be using it.

### DIFF
--- a/core/roslib/include/ros/package.h
+++ b/core/roslib/include/ros/package.h
@@ -29,6 +29,7 @@
 #define ROSLIB_PACKAGE_H
 
 #include <string>
+#include <utility>
 #include <vector>
 #include <map>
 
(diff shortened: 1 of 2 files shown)


core/roslib/include/ros/package.h
```diff
@@ -108,10 +108,18 @@ ROSLIB_DECL bool getAll(V_string& packages);
 ROSLIB_DECL void getPlugins(const std::string& package, const std::string& attribute, V_string& plugins, bool force_recrawl=false);
 
 /**
+ * \brief Call the "rospack plugins" command, eg. "rospack plugins --attrib=<attribute> <package>".  Returns a pair of vector of strings which
+ * are package names and export values respectively. This allows for multiple values in the one package name.
+ */
+ROSLIB_DECL void getPlugins(const std::string& package, const std::string& attribute, V_string& packages, V_string& plugins, bool force_recrawl=false);
```
dirk-thomas: Wouldn't it be better to return a vector of pairs here?

stonier: Yes, I'd prefer a pair also. I was aiming however to keep this change minimal as a working implementation is already there, but hidden, inside the cpp translation unit.

Would you like to push for an upgrade to a vector of pairs?

dirk-thomas: Yes, I think adding a new public function with a vector of pairs would be cleaner. And this function (with two vectors) can stay internal. Can you update the PR that way? Updating only one is ok - I can cherry-pick the changes after one has been merged.

stonier:
- Api using pairs is in.
- Simply reuses the hidden function so there is only one set of worker code in the `.cpp`. I don't think the loss in time in rearranging is significant for anyone in this use case (not likely to be used in a high frequency loop).
- Added some better documentation.

The api is confusing in that the concept of plugins has rather outgrown itself to be a more general export idea propogated by catkin. However, it can't really be overhauled here since rospack itself still uses the plugins terminology. So I've just opted for highlighting more clearly what the name/attribute variables are with an example.

dirk-thomas:
I have created #105 to replace this. I applied some minor style changes and removed the deprecation for the indigo branch.

Thank you for providing this patch!

pavloblindnology:
Hi. Is there Python API for `getPlugins`?

PR #101, opened 2015-12-29
add ability to load tests using dotnames in rosunit

heuristicus:
mirrors changes in ros/ros_comm#722

PR #104, opened 2016-03-01
add ability to load tests using dotnames in rosunit

dirk-thomas:
Replaces #101.

Simplify the selection logic and cleanup new Python code.

### DIFF
--- a/tools/rosunit/src/rosunit/pyunit.py
+++ b/tools/rosunit/src/rosunit/pyunit.py
@@ -57,6 +57,8 @@ def unitrun(package, test_name, test, sysargs=None, coverage_packages=None):
     
     @param package: name of ROS package that is running the test
     @type  package: str
+    @param test: a test case instance or a name resolving to a test case or suite
+    @type  test: unittest.TestCase, or string
     @param coverage_packages: list of Python package to compute coverage results for. Defaults to package
     @type  coverage_packages: [str]
     @param sysargs: (optional) alternate sys.argv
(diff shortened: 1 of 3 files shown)


dirk-thomas:
Cherry picked to jade-devel: 7b8a64dd60a60059846367613a11148e9c92734b

tools/rosunit/src/rosunit/pyunit.py
```diff
@@ -84,7 +86,12 @@ def unitrun(package, test_name, test, sysargs=None, coverage_packages=None):
         start_coverage(coverage_packages)
 
     # create and run unittest suite with our xmllrunner wrapper
-    suite = unittest.TestLoader().loadTestsFromTestCase(test)
+    suite = None
+    if issubclass(test, unittest.TestCase):
+        suite = unittest.TestLoader().loadTestsFromTestCase(test)
+    else:
+        suite = unittest.TestLoader().loadTestsFromName(test)
```
stonier, line 93: This fails. I guess it broke when simplifying the logic in #101 to create PR #104. 

The intention is for a string to pass through the first check into the `loadTestsFromName` call, but it never gets there since the first method throws a `TypeError`.

```
$ python
Python 2.7.6 (default, Mar 22 2014, 22:59:56) 
[GCC 4.8.2] on linux2
Type "help", "copyright", "credits" or "license" for more information.
>>> import unittest
>>> issubclass("foo", unittest.TestCase)
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
TypeError: issubclass() arg 1 must be a class
>>>
```

Not a very intuitive function that. If it's not going to return False, it's going to be a good source of bugs like this.

dirk-thomas: Thank you for reporting this and providing a PR (#121).

stonier:
Thanks @dirk-thomas. Had Korean thanksgiving get in the way and just got back only to find it already done.

dirk-thomas:
No problem. I just wanted to release a new version so went ahead.

PR #107, opened 2016-03-03
[rosbash] Add roscat for indigo as well as jade on #99

wkentaro:


dirk-thomas:
No need for a separate PR. I can cherry-pick the change to `indigo-devel` once it has been merged.

PR #108, opened 2016-03-03
Fixing a few incorrect method calls in junitxml.py

pjreed:
@evenator noticed that a few method calls in junitxml.py were missing parentheses, and one place was incorrectly calling xml() instead of xml_element(), so I fixed them.


dirk-thomas:
Since the tests passed before I think it would be good to add a unit test to ensure there code parts work. Can you please add a small test which covers this?

pjreed:
I've added a few tests that cover the methods I just changed, and I've verified that they fail without those changes.

dirk-thomas:
Thank you very much!

PR #109, opened 2016-03-12
Fix a regression in XML reports introduced by #89

evenator:
The changes in #89 to use an XML framework to generate test reports
introduced a bug caused by variable name collisions in the
_print_error() method. This resolves that bug and corrects the
docstring of the method.

I made this PR against kinetic, but it should be applied to any branch that #89 was merged into.


evenator:
Fixing this uncovered another bug underneath it. I'll remove the [WIP] when it's ready

evenator:
Ok this seems to produce the expected results when I run `rostest roswtf roswtf.test` on kinetic: https://gist.github.com/evenator/9aa19552f6a509fac769

dirk-thomas:
Thank you for the fast patch. Can you please try to add coverage of that code part to the existing test?

evenator:
It doesn't look like there's any test coverage at all on the xmlrunner module, only on junitxml. Am I missing something?

evenator:
Ahh, so they exist, but they're not being run by CMake/catkin. Several tests fail on both the pre- and post- patch, although I think the post-patch failures are whitespaces changes in the output.

evenator:
In addition to the whitespace changes, it seems that it's mangling some poorly-inserted CDATA. I'll try to track down the problem...

evenator:
I fixed the CDATA problem, but the whitespace issues remain. Python's xml library has no pretty-print functionality (minidom does, but it lacks CDATA support as well, so that might take some hacking). Is pretty-printed XML important here, or should I change the tests to pass with non-pretty XML?

Also, what's the best way to have CMake run these tests automatically? I'm not very familiar with python testing in catkin.

dirk-thomas:
The testing part needs some more work but in order to unblock the release of the `ros` repos I have cherry-picked your commit with the fix to the kinetic-devel branch: b40a0b775e44513d1b092dbe741f548d3ee9b185. If that works out I will backport it to Indigo and Jade as well.

Please rebase this PR. I don't know what the right way is and will look at it in the next days again. But a file in the `src` folder should not be used for tests. All testing code should be in the `test` folder.

evenator:
> I have cherry-picked your commit with the fix to the kinetic-devel branch: b40a0b7. If that works out I will backport it to Indigo and Jade as well.

Cool. Thanks! Note that the CDATA tags are still mangled there; you might also want to cherry-pick evenator/ros@d657a653c as well.

> Please rebase this PR.

Will do.

> I don't know what the right way is and will look at it in the next days again.

The main question in my mind is whether pretty-printing the XML is important. AFAIK, the output is perfectly valid, just not very human-readable. So the question becomes, does anyone actually read these XML files? If the answer is yes, the best path forward is probably to bring in lxml, which not only has native pretty-printing, but also handles CDATA properly. The less-fragile CDATA support may be reason enough by itself.

> But a file in the src folder should not be used for tests. All testing code should be in the test folder.

Agreed. Fortunately, it looks like it should be pretty easy to refactor the existing tests out into a separate test script.

dirk-thomas:
The CDATA support was that way before this patch. Before merging the modified CDATA handling I think this certainly needs test coverage.

If not printing the xml with human readable indentation makes the code easier and requires less custom code (like the custom `_serialize_xml` function by using a different library) I would be fine dropping that.

evenator:
The custom `_serialize_xml` is for proper CDATA support, not pretty-printing. As it stands right now, this doesn't do pretty-printing, and if it's not important, I won't pursue it.

When I get a chance to come back to this, I'll rebase, refactor the tests into a separate script, and make sure everything's passing and valid. I'll also have to test it under Python 3.

All in all, I think the whole thing would be cleaner and less fragile with lxml, but I recall there was an objection to pulling in a new dependency.

dirk-thomas:
I understand what the custom serialize implementation is good for. But currently it isn't covered by a test and therefore I am hesitant to merge d657a653cfc0c316a2046838076de3505feee60f. The best way is to make a PR which adds a test in the first commit, CI runs and fails the test, then add a second commit with this patch, and CI confirms that the test passes now.

The pretty-print is not a strict requirement - only a nice to have.

But introducing an extra dependency on this core package is something which should be avoided if possible.

evenator:
It seems that the work and the discussion on this PR have strayed outside of the original request, which was fulfilled when you cherry-picked. I'm closing this and will open new issues/PRs for adding the tests and fixing CDATA.

PR #111, opened 2016-03-17
Avoid putting the rosbuild stacks dir on RPP if it doesn't exist.

mikepurvis:


dirk-thomas:
Sorry for the late response and thank you for the patch.

This will require users to source the setup file (again) after creating the `stacks` folder. That sounds like a reasonable compromise to me in order to keep the environment variables shorter in the common cases.

Issue #112, opened 2016-03-24
rosunit Incorrectly Serializes CDATA


evenator:
rosunit outputs the stdout and stderr using CDATA tags. However, the CDATA tags are incorrectly escaped by the XML serializer like so:

``` xml
<testsuite errors="0" failures="0" name="unittest.suite.TestSuite" tests="1" time="0.000">
<testcase classname="test.test_xmlrunner.TestTest" name="test_foo" time="0.000" />
<system-out><![CDATA[
Foo &gt; Bar
]]></system-out>
<system-err>&lt;![CDATA[
Baz
]]&gt;</system-err></testsuite>
```

Correctly serialized XML would look like:

``` xml
<testsuite errors="0" failures="0" name="unittest.suite.TestSuite" tests="1" time="0.000">
<testcase classname="test.test_xmlrunner.TestTest" name="test_foo" time="0.000" />
<system-out><![CDATA[
Foo > Bar
]]></system-out>
<system-err><![CDATA[
Baz
]]></system-err></testsuite>
```

dirk-thomas:
Closing due to age and inactivity.

PR #113, opened 2016-03-24
Enable rosunit tests

evenator:
There were pre-existing tests for xmlrunner.py in rosunit, but they were in the package itself rather than the test directory. This PR refactors those tests out into the test directory so that catkin runs them automatically and updates the tests to match the current output of the xmlrunner.

Note that this does not address #112: these tests pass even with the current, incorrect serialization of CDATA.


dirk-thomas:
Thank you!

Partially squashed and cherry-picked in: 67e8c57bdd009b18cd2b13717a8ec65296033b85 and e190f82b40ade83009ad5f2b8d225b1e1d097789

PR #118, opened 2016-09-08
Correct spelling of 'rosed'

JacobAMason:
Found this misspelling. There are a couple more instances of 'rossed' in this file, but those may be intentional.


dirk-thomas:
Thanks for the fix. I have cherry-picked your commit to the _kinetic-devel_ branch: e39d114e1a8fc22c8661d1b5572403e0030f63e7

PR #119, opened 2016-09-09
Allow custom classname, testcase name in test_xx_junit_xml.

mikepurvis:
With the default value of "Results", tests which use this code (like the roslaunch check) get a very unhelpful description in the Jenkins layout:

screen shot 2016-09-08 at 10 19 35 pm

By specifying the classname as `roslaunch.RoslaunchCheck` and the case name as, eg `test_my_pkg_check_launch`, it would show as originating from the roslaunch package, and be much more identifiable as to what it is checking (similar to rostests being grouped under `rostest.runner`).

This is a no-function change that is just to enable this improved behaviour in the roslaunch check.

@dirk-thomas

### DIFF
--- a/tools/rosunit/src/rosunit/junitxml.py
+++ b/tools/rosunit/src/rosunit/junitxml.py
@@ -454,7 +454,7 @@ def read_all(filter_=[]):
     return root_result
 
 
-def test_failure_junit_xml(test_name, message, stdout=None, classname="Results"):
+def test_failure_junit_xml(test_name, message, stdout=None, classname="Results", testcasename="test_ran"):
     """
     Generate JUnit XML file for a unary test suite where the test failed
     
@@ -472,7 +472,7 @@ def test_failure_junit_xml(test_name, message, stdout=None, classname="Results")
     testsuite.set('errors', '0')
     testsuite.set('name', test_name)
     testcase = ET.SubElement(testsuite, 'testcase')
-    testcase.set('name', 'test_ran')
+    testcase.set('name', testcasename)
     testcase.set('status', 'run')
     testcase.set('time', '1')
     testcase.set('classname', classname)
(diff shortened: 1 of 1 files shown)


mikepurvis:
There'd be an argument to be made here for `test_name`, `class_name`, `testcase_name`, but I'm easy going on it— would like to get this merged one way or another.

dirk-thomas:
Yes, `class_name` and `testcase_name` would be more readable imo.

Can you please create the PR against `ros_comm` using these new keyword arguments. Are there any other cases which you have in mind which could be updated to use these?

mikepurvis:
Update names. Two additional thoughts:
- Would we consider adding the option to attach stdout to non-failing test?
- Should the `<system-out/>` element by moved inside `<testcase/>`? This seems to be scheme used by nose's XML reports, eg: http://nose2.readthedocs.io/en/latest/plugins/junitxml.html

Will supply PR to ros_comm for roslaunch-check shortly.

dirk-thomas:
> Would we consider adding the option to attach stdout to non-failing test?

Sounds good to me. Can be helpful to know what was happening.

> Should the `<system-out/>` element by moved inside `<testcase/>`?

As long as it is compliant which how it is being parsed, how Jenkins interprets the result and we have the output on a per-test level that sounds good to me too.

mikepurvis:
I think this is good to go now. Example outputs and Jenkins screencaps in https://github.com/ros/ros_comm/pull/897.

dirk-thomas:
Based on you comment (https://github.com/ros/ros_comm/pull/897#issuecomment-247497546) and some testing (afaik Jenkins won't show any `system-out` if the test passed and the location of the `system-out` in the case of the failing test doesn't make a difference it is shown either way) can you please remove the stdout related changes from the PR for now (basically the last commit)?

mikepurvis:
I confirm that Jenkins doesn't show stdout on success. Lame.

I've withdrawn that commit.

dirk-thomas:
Thanks for the patch and iterating on it.

PR #121, opened 2016-09-11
Bugfix to allow loading test cases from name

stonier:
This fixes the problem identified in PR #104. Alternatively could be done by doing the exception handling for `TypeError` if you wish to handle the majority case (class object) first.

### DIFF
--- a/tools/rosunit/src/rosunit/pyunit.py
+++ b/tools/rosunit/src/rosunit/pyunit.py
@@ -87,10 +87,10 @@ def unitrun(package, test_name, test, sysargs=None, coverage_packages=None):
 
     # create and run unittest suite with our xmllrunner wrapper
     suite = None
-    if issubclass(test, unittest.TestCase):
-        suite = unittest.TestLoader().loadTestsFromTestCase(test)
-    else:
+    if isinstance(test, basestring):
         suite = unittest.TestLoader().loadTestsFromName(test)
+    elif issubclass(test, unittest.TestCase): # throws TypeError if issubclass fails
+        suite = unittest.TestLoader().loadTestsFromTestCase(test)
 
     if text_mode:
         result = unittest.TextTestRunner(verbosity=2).run(suite)


stonier:
If this is confirmed, I'll spin up a PR for kinetic and also similar changes over in `ros_comm/rostest/__init__.py`

tools/rosunit/src/rosunit/pyunit.py
```diff
@@ -87,10 +87,10 @@ def unitrun(package, test_name, test, sysargs=None, coverage_packages=None):
 
     # create and run unittest suite with our xmllrunner wrapper
     suite = None
-    if issubclass(test, unittest.TestCase):
-        suite = unittest.TestLoader().loadTestsFromTestCase(test)
-    else:
+    if isinstance(test, basestring):
```
dirk-thomas, line 90: This will raise a `NameError` in Python 3.

tools/rosunit/src/rosunit/pyunit.py
```diff
@@ -87,10 +87,10 @@ def unitrun(package, test_name, test, sysargs=None, coverage_packages=None):
...
         suite = unittest.TestLoader().loadTestsFromName(test)
+    elif issubclass(test, unittest.TestCase): # throws TypeError if issubclass fails
```
dirk-thomas, line 92: Nitpick: please use two spaces before the `#` to comply with PEP8.

tools/rosunit/src/rosunit/pyunit.py
```diff
@@ -87,10 +87,10 @@ def unitrun(package, test_name, test, sysargs=None, coverage_packages=None):
...
+        suite = unittest.TestLoader().loadTestsFromTestCase(test)
```
dirk-thomas, line 93: Now this needs to handle the `else` case somehow.

dirk-thomas:
Thank you for the pull request. I did't notice that the slight change to the logic had that negative side effect. The fix in general looks good beside the comments I added inline.

dirk-thomas:
I created #123 with a much smaller patch. Therefore I will close this as a duplicate.

PR #122, opened 2016-09-16
Allow expected ROS_DISTRO value to be overridden at compile time.

mikepurvis:
@dirk-thomas


dirk-thomas:
Interesting use case :wink: but straight forward patch.

PR #123, opened 2016-09-16
fix check of test type

dirk-thomas:
Replaces #121.


wjwwood (approved):

dirk-thomas:
See 2736a11ad9f851547df9bf25cc949b065aa09c6c for an additional fix since several callers don't pass instances but types.

PR #124, opened 2016-09-19
[indigo] Fix check of test type (#121) (#123)

stonier:
A cherry-pick back to the indigo release branch.


dirk-thomas:
This patch fails a lot of tests. I will backport the patches from the kinetic-devel branch before doing new releases.

PR #125, opened 2016-10-26
Add deduction message type to rostopic pub.

abrzozowski:
This PR is related to http://answers.ros.org/question/246277/command-line-rostopic-pub-why-is-required-specify-a-topic-type/


dirk-thomas:
Thank you!

PR #126, opened 2016-11-07
Added --size option to rosclean purge

kmalhan:
This pull request adds "--size" option to "rosclean purge", which allows user to specify minimum total size in MB to keep when deleting old files. When "--size" option is specified, old log files (based on last modified time) will be deleted until reaches specified minimum total size.

Example Usage
```bash
rosclean purge --size 200
```

### DIFF
--- a/tools/rosclean/src/rosclean/__init__.py
+++ b/tools/rosclean/src/rosclean/__init__.py
@@ -173,15 +173,18 @@ def _rosclean_cmd_purge(args):
             except:
                 print("FAILED to execute command", file=sys.stderr)
         else:
-            print("Purging %s until directory size is at most %d MB."%(label, args.size))
+            files = _sort_file_by_oldest(d)
+            log_size = get_disk_usage(d)
+            if log_size <= args.size * 1024 * 1024:
+                print("Directory size of %s is %d MB which is already below the requested threshold of %d MB."%(label, log_size / 1024 / 1024, args.size))
+                continue
+            print("Purging %s until directory size is at most %d MB (currently %d MB)."%(label, args.size, log_size / 1024 / 1024))
             if not args.y:
                 print("PLEASE BE CAREFUL TO VERIFY THE COMMAND BELOW!")
                 if not _ask("Purge some of old logs in %s"%d):
                     return
-            files = _sort_file_by_oldest(d)
-            log_size = get_disk_usage(d)
             for f in files:
-                if log_size <= args.size * 1000 * 1000:
+                if log_size <= args.size * 1024 * 1024:
                     break
                 path = os.path.join(d, f)
                 log_size -= get_disk_usage(path)
--- a/tools/rosclean/src/rosclean/__init__.py
+++ b/tools/rosclean/src/rosclean/__init__.py
@@ -173,7 +173,7 @@ def _rosclean_cmd_purge(args):
             except:
                 print("FAILED to execute command", file=sys.stderr)
         else:
-            print("Purging %s until directory size is %d MB."%(label, args.size))
+            print("Purging %s until directory size is at most %d MB."%(label, args.size))
             if not args.y:
                 print("PLEASE BE CAREFUL TO VERIFY THE COMMAND BELOW!")
                 if not _ask("Purge some of old logs in %s"%d):
@@ -181,16 +181,15 @@ def _rosclean_cmd_purge(args):
             files = _sort_file_by_oldest(d)
             log_size = get_disk_usage(d)
             for f in files:
-                log_size -= get_disk_usage(os.path.join(d, f))
-                if (log_size >= (args.size * 1000 * 1000) ):
-                    name = d + '/' + f
-                    cmds = [['rm', '-rf', name]]
-                    try:
-                        _call(cmds)
-                    except:
-                        print("FAILED to execute command", file=sys.stderr)
-                else:
+                if (log_size <= (args.size * 1000 * 1000) ):
                     break
+                path = os.path.join(d, f)
+                log_size -= get_disk_usage(path)
+                cmds = [['rm', '-rf', path]]
+                try:
+                    _call(cmds)
+                except:
+                    print("FAILED to execute command", file=sys.stderr)
 
 def rosclean_main(argv=None):
     if argv is None:
@@ -202,7 +201,7 @@ def rosclean_main(argv=None):
     parser_purge = subparsers.add_parser('purge', help='Remove log files')
     parser_purge.set_defaults(func=_rosclean_cmd_purge)
     parser_purge.add_argument('-y', action='store_true', default=False, help='CAUTION: automatically confirms all questions to delete files')
-    parser_purge.add_argument('--size', action='store', default=None, type=int, help='Minimum total size to keep when deleting old files')
+    parser_purge.add_argument('--size', action='store', default=None, type=int, help='Maximum total size in MB to keep when deleting old files')
     args = parser.parse_args(argv[1:])
     args.func(args)
 
--- a/tools/rosclean/src/rosclean/__init__.py
+++ b/tools/rosclean/src/rosclean/__init__.py
@@ -44,6 +44,17 @@ import rospkg
 
 class CleanupException(Exception): pass
 
+def _check_user_input():
+    """
+    Wait for user input and return result
+    @return: ``True`` if user input is ``y``
+    """
+    while 1:
+        input = sys.stdin.readline().strip().lower()
+        if input in ['y', 'n']:
+            break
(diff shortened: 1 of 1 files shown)


kmalhan:
@dirk-thomas Could you provide a feedback to this pull request?

tools/rosclean/src/rosclean/__init__.py
```diff
@@ -66,6 +66,20 @@ def quote(s):
...
+    :param comment: comment, ``str``
+    :return: ``True`` if user responds with y
+    """
+    sys.stdout.write("Okey to perform:\n\n%s\n(y/n)?\n"%comment)
+    while 1:
+        input = sys.stdin.readline().strip().lower()
+        if input in ['y', 'n']:
+            break
+    return (input == 'y')
+
```
dirk-thomas, line 77: The same code lines are present in `_ask_and_call`. Please refactor the code to avoid the code duplication.

tools/rosclean/src/rosclean/__init__.py
```diff
@@ -138,20 +152,68 @@ def get_disk_usage(d):
...
         raise CleanupException("rosclean is not supported on this platform")
 
+def _sort_file_by_oldest(d):
+    """
+    Get files and directories in specified path sorted by last modified time
+    :param d: directory path, ```str```
+    :return:  a list of files and directories sorted by last modified time (old first), ```list```
+    """
+    files = os.listdir(d)
+    files.sort(key=lambda file: os.path.getmtime(os.path.join(d,file)))
```
dirk-thomas: Please avoid using `file` as a variable name since it is a type in Python.

tools/rosclean/src/rosclean/__init__.py
```diff
@@ -138,20 +152,68 @@ def get_disk_usage(d):
...
-                _ask_and_call(cmds)
-        except:
-            print("FAILED to execute command", file=sys.stderr)
+                exe = _ask("Purge some of old logs in %s"%d)
+            else:
+                exe = True
+            if exe:
+                files = _sort_file_by_oldest(d)
+                for file in files:
+                    if _check_delete_file_size(d, file) >= (args.size * 1024 * 1024):
```
dirk-thomas: Currently the algorithm calls `get_disk_usage()` repeatedly. I would suggest to only compute it once. And then for item being considered to be removed only calculate the size of that item. The new remaining size can then be computed from these two in each round of the loop.

tools/rosclean/src/rosclean/__init__.py
```diff
@@ -138,20 +152,68 @@ def get_disk_usage(d):
...
                 print("PLEASE BE CAREFUL TO VERIFY THE COMMAND BELOW!")
-                _ask_and_call(cmds)
-        except:
-            print("FAILED to execute command", file=sys.stderr)
+                exe = _ask("Purge some of old logs in %s"%d)
+            else:
+                exe = True
+            if exe:
+                files = _sort_file_by_oldest(d)
+                for file in files:
```
dirk-thomas: Please avoid using `file` as a variable name since it is a type in Python.

tools/rosclean/src/rosclean/__init__.py
```diff
@@ -138,20 +152,68 @@ def get_disk_usage(d):
...
+            except:
+                print("FAILED to execute command", file=sys.stderr)
+        else:
+            print("Purging %s until directory size is %d MB."%(label, args.size))
+            if not args.y:
                 print("PLEASE BE CAREFUL TO VERIFY THE COMMAND BELOW!")
-                _ask_and_call(cmds)
-        except:
-            print("FAILED to execute command", file=sys.stderr)
+                exe = _ask("Purge some of old logs in %s"%d)
```
dirk-thomas: If the user replies "no" you could simply `return` here which would avoid the `exe` variable as well as the indentation of all of the following code.

tools/rosclean/src/rosclean/__init__.py
```diff
@@ -138,20 +152,68 @@ def get_disk_usage(d):
...
+    :param file: file or directory, ```str```
+    :return: size of directory after purging file, ```int```
+    """
+    d_size = get_disk_usage(d)
+    target = os.path.join(d, file)
+    if os.path.isdir(target):
+        files = os.listdir(target)
+        f_size = 0
+        for f in files:
+            f_size = f_size + os.path.getsize(os.path.join(target, f))
```
dirk-thomas: Can be simplified to `f_size += ...`.

dirk-thomas (commented):
The feature in general looks good. I added a few comments though which should be addressed to clean up the patch.

kmalhan:
Updated to reflect comments.

tools/rosclean/src/rosclean/__init__.py
```diff
@@ -138,20 +155,69 @@ def get_disk_usage(d):
...
+    target = os.path.join(d, f_name)
+    f_size = 0
+    if os.path.isdir(target):
+        for dir_path, dir_name, filenames in os.walk(target):
+            for f in filenames:
+                fp = os.path.join(dir_path, f)
+                f_size += os.path.getsize(fp)
+    else:
+        f_size = os.path.getsize(target)
+    return (log_size - f_size)
```
dirk-thomas: This function doesn't actually care about `log_size`. I suggest removing the argument and simply return `f_size`. The calling code can then do the simple subtraction.

Isn't the function obsolete then since `get_disk_usage(target)` provides the same information?

tools/rosclean/src/rosclean/__init__.py
```diff
@@ -57,15 +68,21 @@ def _ask_and_call(cmds, cwd=None):
...
     return accepted
 
+def _ask(comment):
+    """
+    ask user with provided comment. If user responds with y, return True
+
+    :param comment: comment, ``str``
+    :return: ``True`` if user responds with y
+    """
+    sys.stdout.write("Okey to perform:\n\n%s\n(y/n)?\n"%comment)
```
dirk-thomas: Spelling: "Okey" -> "Okay.

tools/rosclean/src/rosclean/__init__.py
```diff
@@ -57,15 +68,21 @@ def _ask_and_call(cmds, cwd=None):
     def quote(s):
         return '"%s"'%s if ' ' in s else s
     sys.stdout.write("Okay to execute:\n\n%s\n(y/n)?\n"%('\n'.join([' '.join([quote(s) for s in c]) for c in cmds])))
-    while 1:
-        input = sys.stdin.readline().strip().lower()
-        if input in ['y', 'n']:
-            break
-    accepted = input == 'y'
+    accepted = _check_user_input()
```
dirk-thomas: This could be changed to:

```
accepted = _ask('\n'.join([' '.join([quote(s) for s in c]) for c in cmds]))
```

Then `_check_user_input` doesn't need to be separated from `_ask` anymore.

dirk-thomas (commented):
Thank you for the quick update. I have a couple additional comments.

dirk-thomas:
The patch looks good now. I am just wondering if the behavior is what the user expects. The help text says "Minimum total size to keep when deleting old files". The progress message says "Purging %s until directory size is %d MB.". That is kind of inconsistent.

Lets consider the following case where each log subfolder is 3 MB in size and the user passes `-size 10`. I would expect the script to delete one subfolder another until the remaining size is <= 10 MB. Currently the script would stop with 12 MB remaining. If e.g. the last subfolder is 1 GB in size the remaining size would be significantly off from what I expect as a user.

Therefore I would suggest the following:
* Change the help message to "Maximum total size to keep when deleting old files".
* Update the progress message to "Purging %s until directory size is at most %d MB".
* Update the code logic to match, e.g. like this:

```
log_size = get_disk_usage(d)
for f in files:
    if (log_size <= (args.size * 1000 * 1000) ):
        break

    path = os.path.join(d, f)
    log_size -= get_disk_usage(path)
    cmds = [['rm', '-rf', path]]
    try:
        _call(cmds)
    except:
        print("FAILED to execute command", file=sys.stderr)
```

dirk-thomas:
I tried the latest version and ran into the issue that I specified a size in MB but nothing happened. This was due to the fact that 1 MB is 1024 * 1024 B (and not 1000 * 1000). I added a commit fixing this and also checked the size before asking the user the question. If the data size is already lower then the threshold it doesn't need to ask the user.

Can you please briefly check that it still works for you before I go ahead and merge this.

kmalhan:
The changes look good to me and worked on my machine. Please merge.

dirk-thomas:
Thank you for iterating on this!

PR #130, opened 2016-11-25
fix race condition creating folder

dirk-thomas:
Fixes ros/ros_comm#940.


MarcSeebold (approved):
Looks correct to me

dirk-thomas:
Thank you for checking.

PR #131, opened 2016-12-01
[rostopic] Rostopic pub autocomplete's now works when options are provided

javierdiazp:
If a user uses the autocomplete option (pressing the TAB key) while using "rostopic pub" with options, the autocomplete function assumes that the options provided were arguments. This leads the function to provide the wrong arguments to autocomplete.

For example: "rostopic pub --once (TAB)" provides a list of message types instead of a list of topics.

This PR solves this problem by differentiating between options and arguments

### DIFF
--- a/tools/rosbash/rosbash
+++ b/tools/rosbash/rosbash
@@ -776,14 +776,40 @@ function _roscomplete_rostopic {
 										COMPREPLY=($(compgen -W "$opts" -- ${arg}))
 										;;
 								pub)
-										if [[ $COMP_CWORD == 2 ]]; then
+										topic_pos=0
+										type_pos=0
+										msg_pos=0
+										
+										COUNTER=2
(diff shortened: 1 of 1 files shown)


jfabry:
This is joint work of Javier with @OmniSliver (credit where credit is due) as part of a course I am teaching at U. Chile. They have a blog about their work (in Spanish) here: https://javierdiazp.github.io/robotica2/

dirk-thomas:
I tried to apply your patch to the kinetic-devel branch but couldn't get it to work without modifications. The topic name completion worked but not the type name completion. It also didn't handle `-f` (as it does `-r`).

Therefore I created a new PR against kinetic-devel (while maintaining your authorship of the commit) with some modifications of your patch. Please see #136 for the modified patch. It would be great if you (@javierdiazp @jfabry) could try the code and confirm that it works for you.

dirk-thomas:
Closing with #136 being merged.

PR #132, opened 2016-12-01
[rostopic] Rostopic pub autocompletes the topic type

javierdiazp:
Despite the topic type being static, the autocomplete function (pressing TAB) doesn't provide it, but instead provides the whole list of types.

With this PR the autocomplete function uses the given topic name to search the corresponding type to autocomplete. If the topic wasn't declared beforehand, it provides the whole list as always.

Edited:
This solution has another implementation (#125). However, that solution uses awk to parse the topic information and that seems to be slower than this implementation because it only uses bash operations to do the parsing.

### DIFF
--- a/tools/rosbash/rosbash
+++ b/tools/rosbash/rosbash
@@ -780,7 +780,14 @@ function _roscomplete_rostopic {
 												opts=`rostopic list 2> /dev/null`
 												COMPREPLY=($(compgen -W "$opts" -- ${arg}))
 										elif [[ $COMP_CWORD == 3 ]]; then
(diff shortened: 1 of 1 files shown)


jfabry:
This is joint work of Javier with @OmniSliver (credit where credit is due) as part of a course I am teaching at U. Chile. They have a blog about their work (in Spanish) here: https://javierdiazp.github.io/robotica2/

dirk-thomas:
This has already been implemented in #125 against the kinetic-devel branch. The feature might be considered for backporting before the next Jade / Indigo releases.

dirk-thomas:
Please consider to provide a PR against the kinetic-devel branch if you would like to propose a revised implementation.

PR #133, opened 2017-01-12
set correct repository/bugtracker url for rosunit repo

v4hn:
I suppose this got moved at some point
or that part of the package.xml was copied over?

Either way, I just spend 5 minutes looking for this package
in the wrong repository...


dirk-thomas:
Thank you for creating a PR for this and fixing the problem for future users.

PR #134, opened 2017-01-12
don't hide error strings from the reader

v4hn:
Over at MoveIt, we currently have a nondeterministically broken travis (see https://github.com/ros-planning/moveit/issues/311).
The issue at hand makes it harder to understand the source of the problem.

It would be great to see this improved and released soon :)

This except-and-raise pattern hides the actual error messages and leaves
us with a generic string "cannot create test results..." instead.
This makes it much harder to track down bugs.

On the other hand the exception text "[Errno 13] Permission denied: 'foobar'"
alone does not tell you that "foobar" is supposed to be the directory for the
tests results.

Thus I fused both exception texts. Because the error message in case of
missing permissions is (given above) "Permission denied", I removed the
"Please check permissions." hint.


dirk-thomas:

PR #135, opened 2017-01-24
Indigo devel

Wintermute1:


dirk-thomas:
Please add a description what this PR is for. The two branches can't simply be merged into each other.

dirk-thomas:
See #141 for latest backports to Indigo.

PR #136, opened 2017-01-25
rostopic pub autocomplete's works when options are provided

dirk-thomas:
Replaces #131.


dirk-thomas:
@ros/ros_team Please review.

wjwwood (approved):
lgtm, I did not run it locally though.

PR #143, opened 2017-04-14
ensure cwd exists

dirk-thomas:
This can happen when `~/.ros` doesn't exist and a custom `ROS_LOG_DIR` is being used. See ros/ros_comm#1030.


wjwwood (approved):
lgtm

I'm assuming the except is about this issue: http://bugs.python.org/issue13498

dirk-thomas:
> I'm assuming the except is about this issue: http://bugs.python.org/issue13498

No, [Python 2.7](https://docs.python.org/2/library/os.html#os.makedirs) simply doesn't have that keyword argument.

PR #144, opened 2017-04-19
allow arguments in EDITOR env in zsh rosed

sztupi:
in zsh executing
$EDITOR ${arg}
fails if EDITOR has space in it (as zsh escapes the content),
e.g. EDITOR='emacs -nw' becomes 'emacs\ -nw'


wjwwood (approved):
Looks reasonable to me.

dirk-thomas:
Thanks for the patch.

PR #147, opened 2017-05-29
add options in completion for roslaunch to roszsh

kopp:
Some options in the zsh completion for roslaunch were missing.  I added those.

Note: The pull request is for kinetic-devel because I currently use kinetic, but it should probably go to lunar, too.


dirk-thomas:
Thanks. I changed the target branch to be "lunar-devel" since changed are always introduced to the latest development branch first. The patch will then be considered for backporting before the next Kinetic release.

PR #151, opened 2017-08-31
now usable as normal python package.

asmodehn:
Following the same pattern as https://github.com/ros/genpy/pull/86 and https://github.com/ros/genmsg/pull/68


dirk-thomas:
The `lunar-devel` branch has been reintegrated into the `kinetic-devel` branch which automatically closed this PR. Please consider to create a new PR targeting the `kinetic-devel` branch instead. I am sorry for the inconvenience.

PR #152, opened 2017-09-15
Do not print ansi escape sequences if not writing to a tty

spaghetti-:
This fixes https://github.com/ros/ros_comm/issues/1152 and gets us
cleaner output when redirecting rostest to a text file.

### DIFF
--- a/tools/rosunit/src/rosunit/baretest.py
+++ b/tools/rosunit/src/rosunit/baretest.py
@@ -40,6 +40,7 @@ executables. These do not run in a ROS environment.
 from __future__ import print_function
 
 import os
+import sys
 try:
     from cStringIO import StringIO
 except ImportError:
@@ -49,6 +50,7 @@ import time
 import signal
 import subprocess
 import traceback
+import re
 
 import rospkg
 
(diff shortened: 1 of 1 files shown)


dirk-thomas:
Rather than identifying the ascii sequences at the end using a regex would it be easier / cleaner / nicer to not print them in the first place?

spaghetti-:
I think it looks nice when printing to a tty but thats subjective. May also be worth observing that most CI platforms these days support colored output from the tests.

dirk-thomas:
I meant to suggest to conditionally print them with / without ascii characters in the first place (not remove the colorization all together).

spaghetti-:
Right, gotcha, but that involves conditional checks in more lines.

dirk-thomas:
> Right, gotcha, but that involves conditional checks in more lines.

I would recommend to use variables for the color codes and conditionally initialize them once in the beginning. So each line doing the printing wouldn't need to know about the condition.

Anyway I don't mind it either or. If you would like to stick to replacing them from buffer please change the patch to perform the functionality in a function in order to call it from both locations (rather than duplicating the logic).

dirk-thomas:
@spaghetti- I just wanted to check if you are planning on updating this patch based on the comments?

spaghetti-:
@dirk-thomas yes, caught up in work. I'll rebase and update the PR in a coupla days.

dirk-thomas:

PR #153, opened 2017-09-30
use python defined error codes rather than hardcoded integers

mikaelarguedas:
follow up of ros/genpy#89. Use python defined error codes everywhere


tfoote (approved):

dirk-thomas:
Thanks!

PR #159, opened 2017-12-17
Updates to roslib.packages.find_resources

tfoote:
Make the code fully comply with the documented behavior to not search the source tree after a successful find in the binary dir. 

And remove the documentation of a raise on not found as it returns an empty list.

### DIFF
--- a/core/roslib/src/roslib/packages.py
+++ b/core/roslib/src/roslib/packages.py
@@ -510,7 +510,9 @@ def find_resource(pkg, resource_name, filter_fn=None, rospack=None):
     for search_path in search_paths:
         matches.extend(_find_resource(search_path, resource_name, filter_fn=filter_fn))
 
-    matches.extend(_find_resource(pkg_path, resource_name, filter_fn=filter_fn))
+    # Only search ROS_PACKAGE_PATH if not matched in binary build dirs.
+    if not matches:
+        matches.extend(_find_resource(pkg_path, resource_name, filter_fn=filter_fn))
 
     # Uniquify the results, in case we found the same file twice, while keeping order
     unique_matches = []
--- a/core/roslib/src/roslib/packages.py
+++ b/core/roslib/src/roslib/packages.py
@@ -475,7 +475,6 @@ def find_resource(pkg, resource_name, filter_fn=None, rospack=None):
         returns True if the it matches the desired resource, ``fn(str)``
     :param rospack: `rospkg.RosPack` instance to use
     :returns: lists of matching paths for resource within a given scope, ``[str]``
-    :raises: :exc:`rospkg.ResourceNotFound` If package does not exist 
     """
 
     # New resource-location policy in Fuerte, induced by the new catkin 


core/roslib/src/roslib/packages.py
```diff
@@ -475,7 +475,6 @@ def find_resource(pkg, resource_name, filter_fn=None, rospack=None):
         returns True if the it matches the desired resource, ``fn(str)``
     :param rospack: `rospkg.RosPack` instance to use
     :returns: lists of matching paths for resource within a given scope, ``[str]``
-    :raises: :exc:`rospkg.ResourceNotFound` If package does not exist
```
dirk-thomas, line 478: Why is this being removed? It looks to me that the `get_path` call in (old) line 495 can raise this exception (https://github.com/ros-infrastructure/rospkg/blob/bffd0aed87fcbf495361289b850be255bd4e1edd/src/rospkg/rospack.py#L203).

core/roslib/src/roslib/packages.py
```diff
@@ -510,7 +509,9 @@ def find_resource(pkg, resource_name, filter_fn=None, rospack=None):
     for search_path in search_paths:
         matches.extend(_find_resource(search_path, resource_name, filter_fn=filter_fn))
 
-    matches.extend(_find_resource(pkg_path, resource_name, filter_fn=filter_fn))
+    # Only search ROS_PACKAGE_PATH if not matched in binary build dirs.
```
dirk-thomas, line 512: This seems to be the opposite of the (old) lines 499-500. Please describe the use case where the current behavior is a problem and why it should be changed in this way.

dirk-thomas (changes requested):

core/roslib/src/roslib/packages.py
tfoote: Well that's an inconsistency in the API, for if you have a package that exists but the resource name doesn't it returns a `[]` If that other path can throw we can't remove it. I can revert that change. A consistent response would be nice but since this is so legacy we probably shouldn't change anything.

core/roslib/src/roslib/packages.py
tfoote: That comment `in any case, use matches from ros_package_path`

contradicts the doc block: `If the resource is found in
    the binary build directory, it will only return matches in that
    directory; it will not return matches from the ROS_PACKAGE_PATH as
    well in this case.`

And the large comment under the doc block about fuerte and newer behavior. 

I'm debugging a test case where it's failing to launch due to detecting multiple copies of a launch file from different workspaces. It's a CI system creating a dynamic overlay for testing.

Calling `find_resource('package_name', 'amcl.launch')` is giving `
['/tmp/tmpE8gej5/src/package_name/launch/am\
cl.launch', '/opt/custom/share/package_name/launch/amcl.launch']`
Digging in here I saw the contradiction of the doc block and the behavior and fixing that resolved the issue with duplicate responses from different overlays.

core/roslib/src/roslib/packages.py
dirk-thomas: Please revert this change then.

core/roslib/src/roslib/packages.py
dirk-thomas: The current patch still contradicts the remaining comment. So this still needs changes one way or the other.

In general I am less inclined to change the code to match the docblock. Updating the docblock to match the implementation seems much safer.

If you think the behavior should be changed please describe the specific use case in detail in order to reproduce it. From the short comment above I am surprised that `am\ cl.launch` is returned when `amcl.launch` was looked for.

dirk-thomas:
@tfoote I just wanted to check what the status of the PR is?

dirk-thomas:

dirk-thomas:
I will go ahead and delete the branch for now since otherwise I will forget to clean it up in the future if it is not used to create a new PR.

For the record the diff of this PR was:

```
diff --git a/core/roslib/src/roslib/packages.py b/core/roslib/src/roslib/packages.py
index c40267d70..6fb353ef2 100644
--- a/core/roslib/src/roslib/packages.py
+++ b/core/roslib/src/roslib/packages.py
@@ -475,7 +475,6 @@ def find_resource(pkg, resource_name, filter_fn=None, rospack=None):
         returns True if the it matches the desired resource, ``fn(str)``
     :param rospack: `rospkg.RosPack` instance to use
     :returns: lists of matching paths for resource within a given scope, ``[str]``
-    :raises: :exc:`rospkg.ResourceNotFound` If package does not exist 
     """
 
     # New resource-location policy in Fuerte, induced by the new catkin 
@@ -510,7 +509,9 @@ def find_resource(pkg, resource_name, filter_fn=None, rospack=None):
     for search_path in search_paths:
         matches.extend(_find_resource(search_path, resource_name, filter_fn=filter_fn))
 
-    matches.extend(_find_resource(pkg_path, resource_name, filter_fn=filter_fn))
+    # Only search ROS_PACKAGE_PATH if not matched in binary build dirs.
+    if not matches:
+        matches.extend(_find_resource(pkg_path, resource_name, filter_fn=filter_fn))
 
     # Uniquify the results, in case we found the same file twice, while keeping order
     unique_matches = []
```

Issue #161, opened 2018-01-03
Bash completion filters out to many files


nalt:
The function _roscomplete_find for tab completion in rosbash filters out hidden files and folders using `"! -regex ".*/[.].*"`.

If a ".." ends up in the package path - which happens when the user manually adds a relative package path, and also for some workspace configurations using catkin, those paths are filtered out as well. As a result, tab completion finds nothing in the affected packages.

This problem could be avoided by changing the regex to `".*/[.][^.].*"`

dirk-thomas:
Please provide a reproducible example.

nalt:
Set `ROS_PACKAGE_PATH=/home/user/ros/ws/../mypack:/opt/ros/kinetic/share`. Now type `rosrun mypack <tab>`. It does not list anything. The path with .. is filtered out by `"! -regex ".*/[.].*"` in: https://github.com/ros/ros/blob/ba451d78b1482101b239328beae5b8e809b1e272/tools/rosbash/rosbash#L459

One can question whether these `..` should be there. Still, this is a valid path.

Now, set `ROS_PACKAGE_PATH=/home/user/ros/mypack:/opt/ros/kinetic/share`. Now type `rosrun mypack <tab>`. It lists the executables, as expected.

dirk-thomas:
Thank you for the example. Please consider to provide a pull request to support this case.

tfoote:
Another approach would be to resolve the path before running the regex. There are a few different potential approaches here: https://stackoverflow.com/questions/3915040/bash-fish-command-to-print-absolute-path-to-a-file

nalt:
Do you have a preferred solution? All these seem to have their disadvantages, e.g. they work only for one path. `realpath` would do the job, but it may print errors.

dirk-thomas:
I am agnostic to how this is being addressed (as long as it makes your use case work without any drawbacks).

dirk-thomas:
Addressed in #162.

PR #162, opened 2018-01-04
Tab-completion: Do not filter out paths containing ./ or ../

nalt:
Tab complete for rosrun, roscd, rosmsg, etc. did not work if ./ or ../ end up in the results of find. This happens ./ or ../ is used in ROS_PACKAGE_PATH.

See also #161


dirk-thomas:
Thank you for the patch.

In the meantime the `lunar-devel` branch was reintegrated into the `kinetic-devel` branch. Therefore I cherry-picked your commit in 06a3597afed12f6c36b608b24964c24fa41e4199.

PR #163, opened 2018-01-08
Updated XMLTestResult class to provide text output simultaneously

### LINKED ISSUE #110: rosunit should allow simultaneous XML output and text output
Right now we are executing unit tests using make test in our buildfarm. 

Fixing an unexpected failure or a crash on a unit test in the buildfarm is an almost impossible task since we have no logs and the XML output only reports a crash and zero useful logs.

Example of XML output on the case of a crash:

> test [serial_driver_test] did not generate test results
>   File "/usr/lib/python2.7/unittest/case.py", line 331, in run
>     testMethod()
>   File "/opt/ros/indigo/lib/python2.7/dist-packages/rostest/runner.py", line 164, in fn
>     self.assert_(os.path.isfile(test_file), "test [%s] did not generate test results"%test_name)
>   File "/usr/lib/python2.7/unittest/case.py", line 424, in assertTrue
>     raise self.failureException(msg)

It would be very helpful to have alll std output as part of the XML report, if it cannot be made visible in the terminal running the tests.

talhaHavadar:
This commit should resolve #110. But it needs to be tested well since I couldn't test it deeply.

### DIFF
--- a/tools/rosunit/src/rosunit/xmlrunner.py
+++ b/tools/rosunit/src/rosunit/xmlrunner.py
@@ -128,8 +128,8 @@ class _XMLTestResult(unittest.TextTestResult):
 
     """
 
-    def __init__(self, classname):
-        unittest.TextTestResult.__init__(self)
+    def __init__(self, classname, stream=None):
+        unittest.TextTestResult.__init__(self, stream, 1, 1)
         self._test_name = classname
         self._start_time = None
         self._tests = []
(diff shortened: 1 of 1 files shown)


dirk-thomas:
The patch currently fails some of the tests. Please update it to at least pass the current testsuite.

dirk-thomas:

talhaHavadar:
No problem I was facing with problems to run tests its probably related with my local env. I will create a new pr as soon as I fixed the issue.

v-lopez:
@talhaHavadar Any updates on this? Are you looking into fixing the issues, or should I?

talhaHavadar:
Of course you can continue on if you have time. I preferred to change _XMLTestResult's super class to unittest.TextTestResult so we can write test results to console in the same time. I am stuck currently, I couldnt test my code.

PR #168, opened 2018-02-11
Add rosconsole echo

efernandez:
This PR update the auto-complete code for `rosconsole` adding the `echo` command added in https://github.com/ros/ros_comm/pull/1324

### DIFF
--- a/tools/rosbash/rosbash
+++ b/tools/rosbash/rosbash
@@ -957,31 +957,78 @@ function _roscomplete_rosconsole {
     if [[ $COMP_CWORD == 1 ]]; then
         opts="echo get list set"
         COMPREPLY=($(compgen -W "$opts" -- ${arg}))
-    elif [[ $COMP_CWORD == 2 ]]; then
-        case ${COMP_WORDS[1]} in
-            echo|get|set|list)
-                opts=$(rosnode list 2> /dev/null)
-                COMPREPLY=($(compgen -W "$opts" -- ${arg}))
-                ;;
-        esac
-    elif [[ $COMP_CWORD == 3 ]]; then
-        case ${COMP_WORDS[1]} in
-            get|set)
-                opts=$(rosconsole list ${COMP_WORDS[2]} 2> /dev/null)
-                COMPREPLY=($(compgen -W "$opts" -- ${arg}))
-                ;;
-            echo)
-                opts="debug info warn error fatal"
-                COMPREPLY=($(compgen -W "$opts" -- ${arg}))
-                ;;
-        esac
-    elif [[ $COMP_CWORD == 4 ]]; then
-        case ${COMP_WORDS[1]} in
-            set)
-                opts="debug info warn error fatal"
-                COMPREPLY=($(compgen -W "$opts" -- ${arg}))
-                ;;
-        esac
+    elif [[ $COMP_CWORD -ge 2 ]]; then
+        if [[ ${arg} =~ \-.* ]]; then
+            case ${COMP_WORDS[1]} in
+                echo)
+                    opts="-h --help -l --level --nocolor -1 --once --topic -v --verbose"
+                    COMPREPLY=($(compgen -W "$opts" -- ${arg}))
+                    ;;
+            esac
+        else
+            if [[ ${COMP_WORDS[1]} == "echo" ]]; then
+                local prev_arg
+                prev_arg="${COMP_WORDS[$(( $COMP_CWORD - 1 ))]}"
+
+                case ${prev_arg} in
+                    --topic)
+                        opts=$(rostopic list 2> /dev/null)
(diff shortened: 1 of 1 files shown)


efernandez:
FYI @mikepurvis @dirk-thomas

dirk-thomas:
With ros/ros_comm#1324 being merged in the `lunar-devel` branch and this repo not distinguishing between Kinetic and Lunar I am wondering how to proceed with this. I would rather not backport new features into Kinetic but also don't want to branch this repo just for this completion script...

efernandez:
@dirk-thomas Do you think I should check for something like `rosversion` for the pkg and make the completion script compatible with kinetic and lunar?

It'd be easier if you create `lunar-devel` here, but I could try that.

dirk-thomas:
> Do you think I should check for something like rosversion for the pkg and make the completion script compatible with kinetic and lunar?

The completion script could check the ROS distro and only conditionally add the new completion options. While not "great" that would be fine with me to get this working for newer distros.

> It'd be easier if you create `lunar-devel` here, but I could try that.

As mentioned in my previous comment I don't think the overhead of branching (and having to maintain these branches in the future) is justified by the improved completion script.

efernandez:
Rebased and updated checking for `rosversion rospy` >= 1.13.7. `rosconsole echo` was merged right after `rospy` version `1.13.6` was released.

For testing in my local checkout of `ros_comm` I only had to do:
``` diff
diff --git a/clients/rospy/package.xml b/clients/rospy/package.xml
index a56d141..ebc057c 100644
--- a/clients/rospy/package.xml
+++ b/clients/rospy/package.xml
@@ -1,6 +1,6 @@
 <package>
   <name>rospy</name>
-  <version>1.13.6</version>
+  <version>1.13.7</version>
   <description>
     rospy is a pure Python client library for ROS. The rospy client
     API enables Python programmers to quickly interface with ROS <a
```

And the auto-completion works as expected.

The reason I didn't check the ros distro (`rosversion -d`) is because in our software it's always `indigo` regardless of the `ros_comm` version that we use. @mikepurvis might be able to comment more on this if needed.

tools/rosbash/rosbash
```diff
@@ -954,30 +954,98 @@ function _roscomplete_rosconsole {
     COMPREPLY=()
     arg="${COMP_WORDS[COMP_CWORD]}"
 
+    VERSION=$(rosversion rospy)
+    if [[ $? == 1 ]]; then
+        VERSION="0.0.0"
+    fi
+
+    VERSION_ECHO="1.13.7"
+    dpkg --compare-versions ${VERSION_ECHO} le ${VERSION}
```
dirk-thomas, line 963: Can we rely on this to be available across all targeted platforms (e.g. Fedora, macOS)?

efernandez: Good call.

Honestly, I google how to reliably compare versions. I picked this option because it was really simple, but it probably isn't available in other OSs.

I'm open to suggestions. What do you think of the two options here: https://stackoverflow.com/questions/16989598/bash-comparing-version-numbers/24067243 ?

I think `sort -V` should work in most systems nowadays, although some people say it's only available in the GNU sort version. If you think it's better to perform the check w/o `sort -V`, I can implement the other option.

efernandez: Actually, it seems the FreeBSD supports `sort -V`. Do you know if macOS supports it? I don't have a Mac at hand atm.

wjwwood: The man pages are online:

https://developer.apple.com/legacy/library/documentation/Darwin/Reference/ManPages/man1/sort.1.html

I don't see a `-V` option.

wjwwood: Seems like a custom bash function is the most portable way to do it (without using Python):

https://stackoverflow.com/a/4025065/671658

wjwwood: Actually, that online man page might be out-of-date, on my macbook I see the option:

```
% man sort | grep '\-V'
     -V, --version-sort
```

dirk-thomas:
Closing due to inactivity.

tools/rosbash/rosbash
efernandez: @dirk-thomas @wjwwood I'm happy to create a new PR, but I'm not sure if you'd be ok with using `sort -V` or you prefer the approach from https://stackoverflow.com/a/4025065/671658 :thinking:

dirk-thomas: > I'm happy to create a new PR

Absolutely, please make sure to target the default branch (which is `noetic-devel`) first and the patch can be backorder afterwards.

> but I'm not sure if you'd be ok with using sort -V or you prefer the approach from https://stackoverflow.com/a/4025065/671658

As long as it is proven to work on currently supported platforms anything is fine with me.

PR #169, opened 2018-02-14
fix zsh tab completion for symlinks

hannometer:
- add '-L' flag to 'find' command that is used, for example, for tab completion of launch files and package nodes/executables
- building a workspace with catkin tools creates symlinks in a package's libexec folder, in which case the tab completion for the executable with rosrun didn't work
- in rosbash, 'find -L' was already used


dirk-thomas:
Thank you for the patch.

PR #171, opened 2018-02-21
Fix environment variable creation for fish shell

TiagoSMarques:
Fixed compiling error on line 356 off the rosfish script.


dirk-thomas:
The patch looks good to me. Thank you for addressing the problem and making a contribution.

@aclough FYI since you originally created the file in #77.

PR #172, opened 2018-03-13
Fix rosfish

amiller27:
The `rosfish` script here is missing a lot of functionality (and some things just plain didn't work, possibly because fish has changed a lot since this was written).  I've gone and ported much of `rosbash` straight into fish and put it here.  Some things are still not implemented or are not as fast as I would like them to be.  However, the autocomplete for the main commands that I tend to use (and I assume most others tend to use) is working.  I can add the rest of the functionality if it's necessary in order to get this merged.

### DIFF
--- a/tools/rosbash/rosfish
+++ b/tools/rosbash/rosfish
@@ -471,6 +471,60 @@ function _roscomplete_launch
         end
     else
         _roscomplete_search_dir '( -type f -regex .*\.launch$ -o -type f -regex .*\.test$ )'
+        # TODO add extra fancy stuff here for launch file arguments
+    end
+end
+
+function _roscomplete_rosbag
+    set -l command (_ros_command_line)
+    set -l arg $command[-1]
+
+    if test (count $command) = 2
+        set -l opts check compress decompress filter fix help info play record reindex
+        for i in $opts
+            echo $i
+        end
+    else
+        if echo -- $arg | grep -qE -- "--.*"
+            set -l opts
+            switch $command[2]
+            case check
+                set opts --genrules --append --noplugins --help
+            case compress decompress
+                set opts --output-dir --force --quiet --help
+            case filter
+                set opts --print --help
+            case fix
+                set opts --force --noplugins --help
+            case info
+                set opts --yaml --key --freq --help
+            case play
+                set opts --help --quiet --immediate --pause --queue --clock --hz --delay --rate --start --skip-empty --loop --keep-alive --try-future-version --topics --bags
+            case record
+                set opts --help --all --regex --exclude --quiet --output-prefix --output-name --split --size --duration --buffsize --limit --node --bz2
+            case reindex
+                set opts --help --force --quiet --output-dir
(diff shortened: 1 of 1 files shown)


tools/rosbash/rosfish
```diff
@@ -354,5 +788,14 @@ complete -x -c rosrun -a '(_roscomplete_rosrun)'
 complete -x -c roslaunch -a '(_roscomplete_launch)'
 complete -x -c rosed -a '(_roscomplete_file)'
 complete -c rosed -l help --description "prints command info"
-complete -x -c roscp -a '(_roscomplete_file)'
+#complete -x -c roscp -a '(_roscomplete_file)'et -x TEST works
```
dirk-thomas: This change looks weird. Please check why it is commented out and what the appended string is.

amiller27: Sorry, that's because I started working on this before #171.  I commented it out (because that line isn't valid fish, as fixed in #171), the appended string is gibberish which was in there before #171 for some reason.  I'll fix it.

amiller27: The reason I left it commented out is because the implementation of `roscp` here doesn't work (and doesn't look like it should work in fish to me), and I haven't fixed it.  I'll remove the gibberish though.  Has this script (before this pr) been tested by anyone with a recent version of fish?

dirk-thomas:
Since you are already reformatting existing code maybe it makes sense to use a consistent indentation throughout the file?

amiller27:
I agree, it's all 4-space indents now.

dirk-thomas:
Thank you for the patch.

> Some things are still not implemented or are not as fast as I would like them to be.
> I can add the rest of the functionality if it's necessary in order to get this merged.

Would you like to get this merged as-is or continue adding more to this PR ( I don't mind either or).

amiller27:
I have a few more commands I'm working on right now, I'll ping you when those are done.  My guess is that it'll be sometime in the next week, and we can merge it then.

amiller27:
@dirk-thomas This is ready to go.  All the functions are implemented now, and everything that's there works.  There are a couple `TODO`'s still scattered around, but they're minor (in terms of functionality they add) and I'm not going to tackle them in this PR.

dirk-thomas:
> This is ready to go

Since you kept committing more changes I was holding off from merging. If you plan do apply further changes the PR can stay open for longer. Just let me know when you think it should be merged.

amiller27:
Yeah thanks.  I thought it was done and then I discovered some minor bugs/improvements through normal use.  I think I'll sit on this for a little while and see if I come up with anything else, I'll ping you again when I'm done with it.

nim65s:
Hello :)
Any news about this ?
Do you need help or some tests, or anything before merging ?

amiller27:
@dirk-thomas @nim65s This should be ready to merge.  I haven't found any more bugs

dirk-thomas:
Thank you for contributing this patch.

PR #174, opened 2018-03-27
add runtime dependency on VIM as it is the default rosed editor

mikaelarguedas:
`rosed` assumes vim installed if `EDITOR` not set

`rosbash` currently doesn't depend on any editor. But [defaults to vim if no `EDITOR` is specified](https://github.com/ros/ros/blob/9585a2904b245a64fd27064a4932c004e96644d8/tools/rosbash/rosbash#L268).
Currently our docker images don't install vim so rosed is not usable out of the box.

### DIFF
--- a/tools/rosbash/package.xml
+++ b/tools/rosbash/package.xml
@@ -14,4 +14,5 @@
   <buildtool_depend version_gte="0.5.72">catkin</buildtool_depend>
 
   <run_depend>catkin</run_depend>
+  <run_depend>vim</run_depend>
 </package>


tfoote:
I'm not sure that this is really our responsibility, we just delegate to the default EDITOR logic and definitely don't have a dependency on vim itself.

And it's kind of against the goals of keeping a stripped down environment to require installing an editor. This will force vim to be installed even if there is a different valid editor. And part of the goal of having docker images is that they're stripped way down. Others with similar issues: https://stackoverflow.com/questions/30853247/how-to-edit-file-after-i-shell-to-a-docker-container?utm_medium=organic&utm_source=google_rich_qa&utm_campaign=google_rich_qa

mikaelarguedas:
In general I agree that docker images should aim to be small, it just caught me by surprise that by starting with a desktop-full image I couldn't use some rosbash tools out of the box.

I guess my concern is that I would expect users to be able to follow the ROS tutorials directly without having to take extra steps. But apparently the tutorial mentions it: `If this example doesn't work it's probably because you don't have the vim editor installed` so I guess that's fine to close this under the assumption that the additional size outweight the out-of-the-box usability.

dirk-thomas:
> In general I agree that docker images should aim to be small, it just caught me by surprise that by starting with a desktop-full image I couldn't use some rosbash tools out of the box.

Maybe the Docker images should be update to include all these pieces to improve the user experience?

mikaelarguedas:
> Maybe the Docker images should be update to include all these pieces to improve the user experience?

Yes we can add the dependency there.
I guess it depends if we value consistency (docker images being as close as possible to what "installing the packages on a fresh system" provides) or usability (provide a better user experience but diverge from what the packages provide out-of-the-box).

tfoote:
Docker focuses on being a minimal install. It doesn't have many things that are in a "base" install or "fresh" install. There are many things in a "default/fresh" ubuntu install that many of us consider core for useability. (wget curl editors etc)

I would bias toward sticking with the minimal installation since it's trivial to add your own layer on top if you want to add your editor of choice etc. Whereas the other way if we inject it at the lower level it's much more work to remove/rebuild a version without it.

mikaelarguedas:
Sounds good to me, thanks for the feedback, closing this then and leaving the docker images as is

PR #175, opened 2018-04-01
rosrun: replaced `...` with $(...)

cwecht:
According to [shellcheck](https://github.com/koalaman/shellcheck/wiki/Sc2006) this is strongly encouraged.


dirk-thomas:
Thank you for the patch.

PR #176, opened 2018-04-03
rosrun: array is now properly expanded in debug-echo

cwecht:
The previous version printed (uninteded?) only the first entry of the array.


dirk-thomas:

PR #182, opened 2018-06-09
Add verbose output to rosrun if catkin_find fails

tik0:
Signed-off-by: Timo Korthals <tkorthals@cit-ec.uni-bielefeld.de>

### DIFF
--- a/tools/rosbash/scripts/rosrun
+++ b/tools/rosbash/scripts/rosrun
@@ -49,13 +49,9 @@ esac
 
 if [[ -n $CMAKE_PREFIX_PATH ]]; then
   IFS=$'\n'
-  catkin_package_libexec_dirs=($(catkin_find --without-underlays --libexec --share "$1"))
-  if [[ $? -ne 0  ]]; then
-    echo "${catkin_package_libexec_dirs[@]}"
-    exit 3
-  fi
+  catkin_package_libexec_dirs=($(catkin_find --without-underlays --libexec --share "$1" 2> /dev/null))
+  debug "Looking in catkin libexec dirs: $IFS$(catkin_find --without-underlays --libexec --share "$1" 2>&1)"
   unset IFS
-  debug "Looking in catkin libexec dirs: ${catkin_package_libexec_dirs[*]}"
 fi
 pkgdir=$(rospack find "$1")
 debug "Looking in rospack dir: $pkgdir"
--- a/tools/rosbash/scripts/rosrun
+++ b/tools/rosbash/scripts/rosrun
@@ -49,7 +49,11 @@ esac
 
 if [[ -n $CMAKE_PREFIX_PATH ]]; then
   IFS=$'\n'
-  catkin_package_libexec_dirs=($(catkin_find --without-underlays --libexec --share "$1" 2> /dev/null))
+  catkin_package_libexec_dirs=($(catkin_find --without-underlays --libexec --share "$1"))
+  if [[ $? -ne 0  ]]; then
+    echo "${catkin_package_libexec_dirs[@]}"
+    exit 3
+  fi
   unset IFS
   debug "Looking in catkin libexec dirs: ${catkin_package_libexec_dirs[*]}"
 fi


tik0:
This enhancement gives the user a clue why `rosrun` fails sometimes.
Without this enhancement, the user is left with error messages which are not related to the error message caused by catkin_find.
For further investigation see [rosrun map_server map_server couldn't find executable](https://answers.ros.org/question/244237/rosrun-map_server-map_server-couldnt-find-executable/?answer=293662#post-id-293662).

tik0:
Without this enhancement, the user is left with no clue when rosrun fails because of catkin_find.
For further investigation see [rosrun map_server map_server couldn't find executable](https://answers.ros.org/question/244237/rosrun-map_server-map_server-couldnt-find-executable/?answer=293662#post-id-293662)

tools/rosbash/scripts/rosrun
```diff
@@ -49,7 +49,11 @@ esac
 
 if [[ -n $CMAKE_PREFIX_PATH ]]; then
   IFS=$'\n'
-  catkin_package_libexec_dirs=($(catkin_find --without-underlays --libexec --share "$1" 2> /dev/null))
+  catkin_package_libexec_dirs=($(catkin_find --without-underlays --libexec --share "$1"))
```
dirk-thomas: I would expect that any output to `stderr` will now appear on the console? That is not desired since the invocation might succeed but still print some warning to `stderr` (which we want to suppress here).

tik0: Suppressing shouldn't be the standard behavior since the user should be required to have an environment where no application succeeds with errors. Could you elaborate more on that case where `catkin_find` prints errors but the overall invocation of `rosrun` succeeds?

dirk-thomas: If e.g. an installed package results in a warning when its manifest is being parsed (due to non-compliance with naming conventions).

tik0: Ok, but this should be considered as a bug on the user side. Reporting such a warning gives a user the possibility to fix his errors (in that case the naming convention) or report the issue to the package maintainers who do not care about e.g. naming conventions.

However, I suggest that the suppression of errors and warnings should be removed since it has the potential to cause even more confusion. An alternative option would be to store the error message of `catkin_find` and print it, if (and only if) `rosrun` fails because of an empty `catkin_package_libexec_dirs` variable.

tik0: Oh, I just see that my PR also has the suggested alternative behavior.

dirk-thomas: > this should be considered as a bug on the user side.
> I suggest that the suppression of errors and warnings should be removed since it has the potential to cause even more confusion.

Actually it isn't. If the package with the warning is installed as a Debian package the user has no way to address the problem since it is likely not even their own package. That is why this case intentionally discards `stderr` output. If a package the user has control about has a warning like this there are numerous other places where that warning will be visible.

> An alternative option would be to store the error message of `catkin_find` and print it, if (and only if) `rosrun` fails because of an empty `catkin_package_libexec_dirs` variable.

It would certainly be good to show the `stderr` output in case the invocation fails. But as long as it succeeds the `stderr` output should stay suppressed.

> I just see that my PR also has the suggested alternative behavior.

Hm, I just asked that in the beginning:

> I would expect that any output to stderr will now appear on the console?

Just from reading the patch I was assuming `stderr` would not be suppressed and only `stdout` would be captured. Have you tried that specific case?

tik0: Mea culpa, so how about this? The existing debug call also prints out the error information in that case. Btw, piping stderr and stdout to two different variables in one call is not directly possible, so we need two executions of `catkin_find`.

```
  IFS=$'\n'
  catkin_package_libexec_dirs=($(catkin_find --without-underlays --libexec --share "$1" 2> /dev/null))
  debug "Looking in catkin libexec dirs: $IFS$(catkin_find --without-underlays --libexec --share "$1" 2>&1)"
  unset IFS
```

Also, outsourcing `catkin_find --without-underlays --libexec --share "$1"` to a function would make it look cleaner.

dirk-thomas: > we need two executions of `catkin_find`.

That sounds reasonable. (While it technically would be a race I hardly expect the underlying information to change between the invocations. Also the second invocation would only be necessary in case of an error so no overhead for normal operations.)

It would be great if you could update the patch accordingly.

> outsourcing `catkin_find --without-underlays --libexec --share "$1"` to a function would make it look cleaner.

I don't see value in a function for a one-line invocation.

tik0: I've updated the PR accordingly.

tools/rosbash/scripts/rosrun
```diff
@@ -50,8 +50,8 @@ esac
 if [[ -n $CMAKE_PREFIX_PATH ]]; then
   IFS=$'\n'
   catkin_package_libexec_dirs=($(catkin_find --without-underlays --libexec --share "$1" 2> /dev/null))
+  debug "Looking in catkin libexec dirs: $IFS$(catkin_find --without-underlays --libexec --share "$1" 2>&1)"
```
dirk-thomas, line 53: Why should the `stderr` output be redirected to `stdout`?

tik0: This ensures that we do not alter the known behavior of `rosrun` w/o `-d` invokation. As mentioned by you in an [earlier post](https://github.com/ros/ros/pull/182#discussion_r251801568), there might be installed packages which generate warnings, but which is totally OK. If we do not redirect `stderr` to `stdout` and run `rosrun` w/o `-d`, the user sees those warnings, because they are not fetched by the empty `debug` function. However, the new behaviour of `rosrun` w/ `-d` is now, that we see the warnings and errors produced by `catkin_find`, which gives the user more insights why his execution fails.

dirk-thomas:
Thanks for the patch and for iterating on it.

PR #183, opened 2018-07-02
remove executable bit from Python modules

jspricke:


tools/roscreate/src/roscreate/roscreatepkg.py
```diff
@@ -1,3 +1,4 @@
+#!/usr/bin/env python
```
dirk-thomas: The combination of removing the executable flag but adding a shebang line doesn't seem to make sense?

jspricke: Totally agreed, updated.

dirk-thomas:
Thanks.

PR #186, opened 2018-08-01
Fix build issue on Windows

johnsonshih:
Add setenv helper function when built for Windows.
Use portable boost::filesystem to create temporary path in set_env_vars()
Use boost::this_thread::sleep in roslib_caller() when run on Windows


dirk-thomas:

PR #189, opened 2018-08-16
Add missing rospack dependency

paulbovbel:
Needed for roscd and rosrun at the very least.

### DIFF
--- a/tools/rosbash/package.xml
+++ b/tools/rosbash/package.xml
@@ -14,4 +14,5 @@
   <buildtool_depend version_gte="0.5.72">catkin</buildtool_depend>
 
   <run_depend>catkin</run_depend>
+  <run_depend>rospack</run_depend>
 </package>


dirk-thomas:
Thank you for the patch.

gavanderhoorn:
@dirk-thomas: would it be possible to get this released into Kinetic? Or the change backported? According to [status_page/ros_kinetic_default](http://repositories.ros.org/status_page/ros_kinetic_default.html?q=rosbash) the latest version of `rosbash` for Kinetic is `1.14.4` which doesn't seem to include this change.

I have a few environments which start from scratch and they're showing `rospack: no such command or filename` errors without this.

dirk-thomas:
Released version 1.14.6 into Kinetic: ros/rosdistro/pull/21494

gavanderhoorn:
Thanks. This simplifies things significantly.

PR #190, opened 2018-08-20
rosunit python 3 support

ablakey:
Address ros#158 by explicitly decoding to utf-8 before writing to stream.

This is a typical Python 2 -> 3 issue. Before this fix, Python 2 would return a `str` and Python 3 would return `bytes`. `bytes` is not supported for writing to streams like stdout in Python 3.

This fix explicitly decodes to utf-8 before writing, so in Python 2 we're writing a `unicode` object and in Python 3 we're writing a `str` object.

### DIFF
--- a/tools/rosunit/src/rosunit/xmlrunner.py
+++ b/tools/rosunit/src/rosunit/xmlrunner.py
@@ -199,7 +199,8 @@ class _XMLTestResult(unittest.TestResult):
         output and standard error streams must be passed in.a
 
         """
-        stream.write(ET.tostring(self.xml(time_taken, out, err).getroot(), encoding='utf-8', method='xml'))
+        root = self.xml(time_taken, out, err).getroot()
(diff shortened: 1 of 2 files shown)


dirk-thomas:
Please avoid any unrelated whitespace changes.

ablakey:
Sorry. My text editor fixes those mistakes automatically.  Reverted.

dirk-thomas:
> Sorry. My text editor fixes those mistakes automatically. Reverted.

I know it is annoying :disappointed: Once we get close to Indigo being EOLed we hopefully have only a single branch to maintain and can cleanup a lot of the code (without having to worry about backporting patches).

Thanks for the patch.

paulbovbel:
Can I request a tagged release that includes this patch please?

dirk-thomas:
It will likely be January until I get to do ROS 1 patch releases.

PR #194, opened 2019-01-22
use _MSC_VER for ROS_FORCE_INLINE

kejxu:
unify the use of macros:
use the `_MSC_VER` macro instead of the `MSVC` macro just like [line 52](https://github.com/ros/ros/blob/kinetic-devel/core/roslib/include/ros/package.h#L52) in the same file


dirk-thomas:
Thanks for the patch.

PR #196, opened 2019-01-22
sync rosbuild.bat template with rosbuild.sh template

kejxu:
update `rosbuild.bat.em` to follow the same logic as in [`rosbuild.sh.em`](https://github.com/ros/ros/blob/kinetic-devel/core/rosbuild/env-hooks/10.rosbuild.sh.em#L8):
use `CATKIN_ENV_HOOK_WORKSPACE` if it's defined


dirk-thomas:
Thanks for the patch.

PR #197, opened 2019-01-22
special handle python script without extension on Windows

kejxu:
python scripts in ROS tend to omit `.py` extension since they could become executable by adding a shebang line:
```
#!/usr/bin/env python
```
in linux environment

special handle this case in Windows environment

### DIFF
--- a/core/roslib/src/roslib/packages.py
+++ b/core/roslib/src/roslib/packages.py
@@ -406,7 +406,7 @@ def _executable_filter(test_path):
     # Python scripts in ROS tend to omit .py extension since they could become executable
     # by adding a shebang line (#!/usr/bin/env python) in Linux environments
     # special handle this case in Windows environment
-    if os.name == 'nt' and os.path.splitext(test_path)[1] in ['.py', '']:
+    if os.name == 'nt' and os.path.splitext(test_path)[1].lower() in ['.py', '']:
         flags = stat.S_IRUSR
     return (s.st_mode & flags) == flags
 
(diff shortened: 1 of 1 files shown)


core/roslib/src/roslib/packages.py
```diff
@@ -402,8 +402,13 @@ def find_node(pkg, node_type, rospack=None):
...
     s = os.stat(test_path)
     flags = stat.S_IRUSR | stat.S_IXUSR
-    if os.name == 'nt' and os.path.splitext(test_path)[1] == '.py':
-        flags = stat.S_IRUSR
+    if os.name == 'nt':
+        _, ext = os.path.splitext(test_path)
+        # python scripts in ROS tend to omit .py extension since they could become executable
+        # by adding a shebang line (#!/usr/bin/env python) in linux environment
+        # special handle this case in Windows environment
+        if ext.lower() in ['.py', '']:
```
dirk-thomas: Is the `.lower` really necessary? I have never encountered `.PY` as a file extension.

dirk-thomas (commented):
The diff could probably collapsed to just change:

> == '.py':

into

> in ['.py', '']

(plus the additional comment).

core/roslib/src/roslib/packages.py
kejxu: we think it wouldn't hurt, there is no guarantee whether anyone would actually use `.PY` as the extension. Meantime, `.lower` should be a quite simple implementation, so it should not hurt complexity or efficiency.

kejxu:
> The diff could probably collapsed to just change:
> 
> > == '.py':
> 
> into
> 
> > in ['.py', '']
> 
> (plus the additional comment).

will change to the suggested change

dirk-thomas:
Thank you for the patch.

PR #198, opened 2019-01-22
enable rosclean on Windows

kejxu:
enable rosclean on Windows by:
- implementing [`get_disk_usage`](https://github.com/ros/ros/blob/kinetic-devel/tools/rosclean/src/rosclean/__init__.py#L130) on Windows
- replace `rm` with `rd` when on Windows


dirk-thomas:
Thanks for the patch.

PR #199, opened 2019-01-22
fall back to default value if modules are not available

kejxu:
`fcntl` is not available on Windows, leading to exception in [`terminal_width()`](https://github.com/ros/ros/blob/kinetic-devel/tools/rosmake/src/rosmake/engine.py#L251)

while it sounds like a Windows-specific issue, yet we are trying to resolve it in a generic way:
if any of the modules (`struct`, `fcntl`, `termios`) are not available, fall back to default value for terminal width


dirk-thomas:
Thanks for the patch.

PR #200, opened 2019-01-22
add ${PYTHON_EXECUTABLE} to ROSUNIT_EXE

kejxu:
python scripts could be directly executed on Linux thanks to the help of the shebang line, so most of the times the `${PYTHON_EXECUTABLE}` prefix is ignored; however, python scripts are not executable on Windows (without extra setup), so the prefix is needed (it wouldn't hurt)


dirk-thomas:
Thanks for the patch.

PR #208, opened 2019-02-25
add rosrun.bat

kejxu:
follow logic in existing [rosrun script](https://github.com/ros/ros/blob/kinetic-devel/tools/rosbash/scripts/rosrun)

special handle execution of Python scripts by executing scripts with the Python interpreter.

### DIFF
--- a/tools/rosbash/CMakeLists.txt
+++ b/tools/rosbash/CMakeLists.txt
@@ -14,9 +14,16 @@ else()
 endif()
 
 # making toplevel forward script for bash script rosrun
-set(BASH_SCRIPT ${CMAKE_CURRENT_SOURCE_DIR}/scripts/rosrun)
-configure_file(${catkin_EXTRAS_DIR}/templates/script.bash.in
-  ${CATKIN_DEVEL_PREFIX}/bin/rosrun
-  @ONLY)
+if(NOT WIN32)
+  set(BASH_SCRIPT ${CMAKE_CURRENT_SOURCE_DIR}/scripts/rosrun)
+  configure_file(${catkin_EXTRAS_DIR}/templates/script.bash.in
+    ${CATKIN_DEVEL_PREFIX}/bin/rosrun
+    @ONLY)
+else()
+  set(BAT_SCRIPT ${CMAKE_CURRENT_SOURCE_DIR}/scripts/rosrun.bat)
+  configure_file(${catkin_EXTRAS_DIR}/templates/script.bat.in
+    ${CATKIN_DEVEL_PREFIX}/bin/rosrun.bat
+    @ONLY)
+endif()
 
 catkin_add_env_hooks(15.rosbash SHELLS bash fish tcsh zsh DIRECTORY ${CMAKE_CURRENT_SOURCE_DIR}/env-hooks)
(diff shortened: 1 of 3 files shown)


tools/rosbash/CMakeLists.txt
```diff
@@ -5,13 +5,25 @@ catkin_package()
...
-  ${CATKIN_DEVEL_PREFIX}/bin/rosrun
-  @ONLY)
+if(NOT WIN32)
+  set(BASH_SCRIPT ${CMAKE_CURRENT_SOURCE_DIR}/scripts/rosrun)
+  configure_file(${catkin_EXTRAS_DIR}/templates/script.bash.in
+    ${CATKIN_DEVEL_PREFIX}/bin/rosrun
+    @ONLY)
+else()
+  set(BAT_SCRIPT ${CMAKE_CURRENT_SOURCE_DIR}/scripts/rosrun.bat)
+  configure_file(${catkin_EXTRAS_DIR}/templates/script.bat.in
```
dirk-thomas, line 24: This will require a to-be-released version of `catkin`. Once `catkin` has been released the `rosbash` package needs to state a version dependency in the package manifest.

kejxu: that makes sense! Totally forgot this change in `rosbash` would have dependency on `catkin`

2 ways I can think of to work around this:
- do we have any release plan for `catkin` in sight? (most Windows-specific changes are either merged or in pull request state already)
- would it be a good idea to Python-ize `rosrun` (or `rosbash` in general) to make the cli become (Python script) executables?

dirk-thomas: > do we have any release plan for `catkin` in sight?

I am planning to do a release after ros/catkin#978 and ros/catkin#993 have been merged.

> would it be a good idea to Python-ize `rosrun` (or `rosbash` in general) to make the cli become (Python script) executables?

I would suggest to not do this kind of rewrite for now but focus on getting the Windows support in. Such a drastic change should be very carefully considered and requires extensive testing. I would certainly not be in favor of releasing such a change into an already existing ROS distro but only consider it for future ROS distros (e.g. Noetic in May 2020).

kejxu: > I am planning to do a release after [ros/catkin#978](https://github.com/ros/catkin/pull/978) and [ros/catkin#993](https://github.com/ros/catkin/pull/993) have been merged.

good to know! can I translate that to a version requirement of `0.7.15` for `catkin`? 
 
> Such a drastic change should be very carefully considered and requires extensive testing. I would certainly not be in favor of releasing such a change into an already existing ROS distro but only consider it for future ROS distros (e.g. Noetic in May 2020).

thanks for the explanation! given it would be a change with huge impact, highly appreciate the carefulness =)

dirk-thomas: > can I translate that to a version requirement of `0.7.15` for `catkin`?

Yes, that is the new version of `catkin` (which I just tagged). Please update the version number here accordingly so that this can be merged:

https://github.com/ros/ros/blob/9c92fa4e6522fa250afcad9fa37cd0ccb47c5b75/tools/rosbash/package.xml#L14

kejxu: thanks for pointing out the place to add the version gate, have updated `package.xml`

dirk-thomas:
Thanks for the patch.

PR #210, opened 2019-03-05
add roscd, rosls support

kejxu:
add `roscd`, `rosls` support by following the logic in [ros/rosbash](https://github.com/ros/ros/blob/kinetic-devel/tools/rosbash/rosbash)

also add Python implementation for `_ros_location_find` and `_ros_decode_path` on Windows to remove dependency on Linux shell features


tools/rosbash/scripts/findpath.py
```diff
@@ -0,0 +1,76 @@
...
+# on success with arguments print result_path or Error: error message
+def findpathmain(argv):
+    reldir = ''
+    parameters = os.path.normpath(argv[0]).split(os.path.sep)
+    package_name = parameters[0]
+    if len(parameters) > 1:
+        reldir = os.path.sep.join(parameters[1:len(parameters)])
+    else:
+        if len(argv) <2 or argv[1] != 'forceeval':
+            print (ERROR_PREFIX + package_name)
```
dirk-thomas: Please remove the space before the open parenthesis. The error message `Error: <pkgname>` doesn't seem very helpful. Also error messages should be printed to `stderr`.

Same below.

tools/rosbash/CMakeLists.txt
```diff
@@ -8,6 +8,9 @@ install(FILES rosbash rosfish rostcsh roszsh
 if (WIN32)
   install(PROGRAMS scripts/rosrun.bat
     DESTINATION ${CATKIN_GLOBAL_BIN_DESTINATION})
+
+  install(PROGRAMS  scripts/findpath.py scripts/roscd.bat scripts/rosls.bat
```
dirk-thomas: The filename `findpath` sounds very generic. How about renaming it to `rosfindpath`?

Nitpick: please remove one of the two spaces after `PROGRAMS`.

tools/rosbash/scripts/findpath.py
kejxu: thanks for pointing this out, updated to print to `stderr`

tools/rosbash/CMakeLists.txt
kejxu: thanks for the name suggestion, updated to `rosfindpath`, removed extra space

dirk-thomas:
Thanks for the patch.

PR #211, opened 2019-03-08
rosbash: exclude files when completing roslaunch args

christophebedard:
This changes roslaunch-complete so that it excludes files/directories when completing launch file args.

I stumbled upon this `FIXME` while looking at ros/ros_comm#1300 and thought I'd submit this.

Here's a quick example. Previously:
```
$ source devel/setup.bash
$ roslaunch tracecompass_ros_testcases pub_sub.launch <tab>
build//              latching             num_pubs             src_nobuild//        .ycm_extra_conf.pyc
devel//              logs//               src//                .ycm_extra_conf.py
```

Now:
```
$ source devel/setup.bash
$ roslaunch tracecompass_ros_testcases pub_sub.launch <tab>
latching  num_pubs
```

I tested to make sure it doesn't affect anything else, but I'd appreciate if someone else could confirm.


dirk-thomas:
Thanks for the patch.

PR #213, opened 2019-03-29
correct findpath script name in batch scripts

kejxu:
`findpath.py` was renamed to `rosfindpath.py` during previous upstreaming processes, and the batch scripts were not changed to reference the new name.

special thanks to @bakerhillpins for reporting this issue in https://github.com/ms-iot/ROSOnWindows/issues/92

not related to this pull request, but one of the open tasks in our backlog is to implement an Azure-pipeline-based CI for Windows, so issues like this would be caught earlier. My apologies for the bug.

### DIFF
--- a/tools/rosbash/scripts/roscd.bat
+++ b/tools/rosbash/scripts/roscd.bat
@@ -22,7 +22,7 @@ if "%1" equ "" (
     exit /b 1
 )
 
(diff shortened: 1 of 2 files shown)


dirk-thomas:
Thanks for the fix.

PR #217, opened 2019-05-16
Feature/zsh launch args

venomyeah:
roslaunch arguments auto-completion feature ported from bash script to zsh script


fmauch:
Awesome! I've just tested it and on melodic those warnings about xacro order fire in:

```bash
$ roslaunch ur_description ur10_upload.launch xacro: in-order processing became default in ROS Melodic. You can drop the option.
limited
```
Completion still works, though.

dirk-thomas:
Thanks for the contribution.

Issue #192, opened 2018-09-04
rosbash: "find: paths must precede expression" on auto-complete attempt of rosmsg (with msg files in the CWD)


gavanderhoorn:
To reproduce:

```
docker run -it --rm ros:melodic
source /opt/ros/melodic/setup.bash
roscd sensor_msgs/msg
rosmsg show sen<tab>
```

error returned then is:

```
find: paths must precede expression: `CameraInfo.msg'
find: possible unquoted pattern after predicate `-name'?
```

I've tried to understand what is going on in the auto-complete hook for `rosmsg`, but I've never had the pleasure of working with `compgen` so it's all black magic to me. It seems like the problem *could be* somewhere around:

https://github.com/ros/ros/blob/fd860146ff14efc425ef020ce570b3ba1e950772/tools/rosbash/rosbash#L902-L905

gavanderhoorn:
I have a hunch that something like [this](https://stackoverflow.com/questions/27922449) is happening.

jexner:
@gavanderhoorn Yes, that was the actual issue. #218 provides a fix.

dirk-thomas:
Addressed by #218.

PR #218, opened 2019-05-16
Prevent filename glob expansion in _msg_opts in rosbash

jexner:
The `find` argument glob is not properly quoted resulting in bash filename
expansion. This leads to incorrect `find` calls.

This fixes issue #192.


jexner:
While looking into this issue I stumbled upon two definitions of the function `_msg_opts` in the file `rosbash`. This seems a bit odd. Is this intended?

dirk-thomas:
Thanks for the patch.

Issue #195, opened 2019-01-22
Please add a license to this repo


mbrukman:
Could you please add an explicit `LICENSE` file to the repo so that it's clear under what terms the content is provided, and under what terms user contributions are licensed?

Per [GitHub docs on licensing](https://help.github.com/articles/licensing-a-repository/#choosing-the-right-license):

> [...] without a license, the default copyright laws apply, meaning that you retain all rights to your source code and no one may reproduce, distribute, or create derivative works from your work. If you're creating an open source project, we strongly encourage you to include an open source license.

Thanks!

dirk-thomas:
Each package in this repository specifies it's license in the `package.xml` file already.

mbrukman:
Thanks for the info! Are they all the same or different (cursory investigation finds a bunch of BSD-licensed modules)? If they're all the same, adding a `LICENSE` file it makes it easier for viewers on GitHub to see what the license directly from the repo overview, thanks to this feature: https://github.blog/2016-09-21-license-now-displayed-on-repository-overview/ so this would be a welcome addition to clarify the licensing.

dirk-thomas:
If someone could check all of them and post a summary here that would be great.

gavanderhoorn:
A quick check of all `package.xml`s shows:

```
mk: bsd
ros: bsd
rosbash: bsd
rosboost_cfg: bsd
rosbuild: bsd
rosclean: bsd
roscreate: bsd
roslang: bsd
roslib: bsd
rosmake: bsd
rosunit: bsd
```

amilcarlucas:
I also need this. But I can not do a PR for legal reasons. :(

luzpaz:
bump

amilcarlucas:
Fixed by #222 . Can be closed now.

PR #222, opened 2019-06-24
Add missing license information

amilcarlucas:
This fixes issue https://github.com/ros/ros/issues/195

### DIFF
--- a/LICENSE
+++ b/LICENSE
@@ -2,16 +2,15 @@ Redistribution and use in source and binary forms, with or without
 modification, are permitted provided that the following conditions
 are met:
 
-1. Redistributions of source code must retain the above copyright
+ * Redistributions of source code must retain the above copyright
    notice, this list of conditions and the following disclaimer.
-2. Redistributions in binary form must reproduce the above copyright
-   notice, this list of conditions and the following disclaimer in
-   the documentation and/or other materials provided with the
-   distribution.
-3. Neither the name Open Source Robotics Foundation nor the names of
-   its contributors may be used to endorse or promote products
-   derived from this software without specific prior written
-   permission.
+ * Redistributions in binary form must reproduce the above
+   copyright notice, this list of conditions and the following
+   disclaimer in the documentation and/or other materials provided
+   with the distribution.
+ * Neither the name of copyright holder nor the names of its
+   contributors may be used to endorse or promote products derived
+   from this software without specific prior written permission.
 
 THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS
 "AS IS" AND ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT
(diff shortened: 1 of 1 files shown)


gavanderhoorn:
Where did you find (L)GPL code in this repository?

amilcarlucas:
The license in the PR is copied from https://github.com/mavlink/mavros/
This repo seams to only contain BSD code, but according to the guidelines from ROS all code should be released under the three licenses.
Did I misunderstood it ?

gavanderhoorn:
> This repo seams to only contain BSD code, 

So there is only one license file needed.

> but according to the guidelines from ROS all code should be released under the three licenses.

which guidelines?

amilcarlucas:
According to http://wiki.ros.org/DevelopersGuide the ROS core is BSD, but because other (non-BSD) software uses it, it needs to be released under multiple licenses. Ask the person who wrote those paragraphs for more information, if you need it.

gavanderhoorn:
Could you clarify which parts of that page you believe state that "it needs to be released under multiple licenses"? I have a hard time finding that.

Note that the DevelopersGuide is a very old page and mostly refers to how things were done in the 2012 / WillowGarage days.

In any case: under [Licensing](http://wiki.ros.org/DevelopersGuide#Licensing), the following is written:

> - The preferred license for the project is the BSD license. Whenever possible, new code should be licensed under BSD. All ROS core code is licensed BSD. 

The `ros/ros` repository contains core libraries and tools, and all packages are licensed BSD.

> - Whenever possible, each ROS package is governed by a single license.

Seems to hold here.

> - Common special case: BSD-licensed code (e.g., the ROS core) is used in a package that also uses GPL-licensed code. To comply with the GPL, the BSD-licensed code is multiply-licensed under BSD and GPL, with the user given the choice of which license to use. This is not a big deal, because the GPL simply adds restrictions that are not in BSD. 

This says that *if* there is code that is GPL licensed, *then* the GPL license must be included and/or dual-licensing is possible.

There is no GPL code in this repository, so this does not seem to apply.

Finally: the "We rigorously obey license terms on third-party software that we use." section states that *if* GPL licensed code is part of a package, *then* you must also use GPL.

But again, there is no GPL code here, so thi doesn't seem to apply.

---

Afaict, that page does not state anywhere that the packages in this repository -- or ROS packages in general -- *must* be dual (or multi-) licensed. In fact, it states:

> The ROS packaging and communication system allows for fine-grained licensing. Because nodes communicate via ROS messages, code from multiple nodes is not linked together. Thus the package provides a kind of "license boundary." 

If my package is GPL, yours is BSD, but we only communicate using messages/services and actions, then the license does not carry over to (or "contaminate") yours.

---

I'll defer to the maintainers here -- as I'm not one of them -- but afaict, only a single BSD license text would need to be added here in the root of the repository. As there is no GPL, LGPL or any other code under other licenses (or at least: that is what the individual package manifests imply), it would appear there is no need for those license texts, nor the mention in the readme.

amilcarlucas:
My fault, I guess I read that guideline from mavros, and confused it with ROS. Corrected now.

amilcarlucas:
Can this be merged now?

dirk-thomas:
> Can this be merged now?

Unfortunately not in the current state.

* Please name the file `LICENSE` (without the `-BSD` suffix) since it is the filename used by convention and e.g. also interpreted by GitHub.
* Also please use the exact content as in other repos (e.g. in [ros/catkin](https://github.com/ros/catkin/blob/321c0fa12360aa6e12c6c780856340f79d3704a5/LICENSE)) without adding a custom copyright line (which by the way is incorrect for this repo).

amilcarlucas:
For legal reasons I need to know who is the copyright holder. I will remove that line from the file.
But who is the copyright holder ?

dirk-thomas:
> who is the copyright holder ?

That is stated for each source file in the file header.

amilcarlucas:
I searched for every copyright notice on every source file and came up with this.

LICENSE
```diff
@@ -0,0 +1,31 @@
+Copyright (c) 2008-2010, 2012. Willow Garage, Inc. All rights reserved.
+Copyright 2010 Kitware, Inc.
+Copyright (C) 2006 Enrico Scholz <enrico.scholz@informatik.tu-chemnitz.de>
```
dirk-thomas: Please remove these lines again since this is unmaintainable.

dirk-thomas: @amilcarlucas Friendly ping.

amilcarlucas:
Done, it can be merged

dirk-thomas:
Thanks for the patch.

I updated the file to match the exact same one in the `catkin` repo. The changes are only formatting / line wrapping.

LICENSE
abuibaid: Hey @dirk-thomas , just wondering why are you saying it is unmaintainable and therefore these lines need to be cleared out while through the code the copyright holder is still there?

I'm asking because I'm publishing code as open source, and for that we need to know the copyright notice of ROS package!

PR #224, opened 2019-07-01
fix rosclean, use rd on Windows

kejxu:
copy fix from https://github.com/ros/ros/blob/kinetic-devel/tools/rosclean/src/rosclean/__init__.py#L218, use `rd /s /q` on Windows

thanks for root-causing this issue, @c650

this addresses https://github.com/ms-iot/ROSOnWindows/issues/111


dirk-thomas:
Thanks for the patch.

PR #225, opened 2019-07-02
[roslib] Symlink search for roslib

ViktorWalter:
Allowing for finding resources symbolically linked in packages. We needed this for using `roslaunch` with directories linked from other locations, but it will likely be useful in general.


dirk-thomas:
Thanks for the patch.

PR #232, opened 2020-01-22
Include FreeBSD in platform-dependent argument for `find -perm`

dodsonmg:
FreeBSD doesn't support `find -perm /111`.  Like MacOS (Darwin), it supports `find -perm +111`.

This PR adds FreeBSD to the existing conditional statements that check for Darwin before deciding whether to use `/111` or `+111`.

This is related to issues #33 and #37.


dirk-thomas:
Thanks for the fix.

PR #233, opened 2020-01-23
[noetic] Rosrun uses heuristic to pick python devel-space relay script

sloretz:
This is to support ros/catkin#1044 .That PR adds a devel-space relay script for installed python executables so their shebang can be rewritten to match `ROS_PYTHON_VERSION`. It has a drawback that `rosrun` will warn about non-unique executables because it sees one in the devel space and one in src space. This PR adds an exception to that warning.

The first commit is a slight refactoring that should not change `rosrun`'s behavior.

1. 6773140 makes the script use `pkg_name` and `file_name` instead of `$1` and `$2` because it is confusing to read when adding a function that also uses arguments.
1. 61d017c  makes `rosrun` pick the first executable if there are exactly two and one came from the `--libexec` list and another came from the `--share` list.

### DIFF
--- a/tools/rosbash/scripts/rosrun
+++ b/tools/rosbash/scripts/rosrun
@@ -50,6 +50,21 @@ esac
 pkg_name="$1"
 file_name="$2"
 
+function inonedir() {
+  exe=$1
+  shift
+  list_of_dirs=("$@")
+  for location in "${list_of_dirs[@]}";
+  do
+    if [[ "$exe" = "$location"* ]] ; then
+      # exe path starts with $location
+      echo "yes"
+      return
+    fi
+  done
+  echo "no"
+}
+
 if [[ -n $CMAKE_PREFIX_PATH ]]; then
   _rosrun_IFS="$IFS"
   IFS=$'\n'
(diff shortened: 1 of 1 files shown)


dirk-thomas:
> 06a2805 splits the `catkin_find` invokaction into two: once with `--libexec` and once with `--share`.

This will slow down the command since Python needs to be invoked another time. Since this will affect every invocation of `rosrun` I am not sure that is ok. The alternative would be to leave the single invocation as is and only if it returned 2 matches query `catkin_find` once again (with only one of the two arguments). That would only increase the runtime when there are actually two matches which should be much less common.

sloretz:
> This will slow down the command since Python needs to be invoked another time. Since this will affect every invocation of rosrun I am not sure that is ok.

It does seem to slow it down; I'll implement an alternative. The average run time of 6 runs on my machine (bailing before exec) is 0.498s before 06a2805, and 0.781s after.

sloretz:
> The alternative would be to leave the single invocation as is and only if it returned 2 matches query catkin_find once again (with only one of the two arguments). That would only increase the runtime when there are actually two matches which should be much less common.

Implemented in ~~35c874f~~ 61d017c

sloretz:
@dirk-thomas mind having another look at this PR? I assume this will need a `noetic-devel` branch since it's only purpose is to support ros/catkin#1044 which has been merged already on catkin's `noetic-devel` branch.

dirk-thomas (approved):
LGTM

Will do the branching and merging when we are about to release this repo.

sloretz:
@ros-pull-request-builder retest this please

dirk-thomas:
@ros-pull-request-builder retest this please

dirk-thomas:
Moved the first commit into a separate PR: #239.

sloretz:
Looks like Noetic PR job passed :tada:

sloretz:
Since this is a companion to ros/catkin#1044 which was merged onto the noetic-devel branch, I think this PR might need to be re-targeted. @dirk-thomas mind creating a `noetic-devel` branch?

dirk-thomas:
> I think this PR might need to be re-targeted.

Yes, as the `[noetic]` prefix in the title indicates.

> mind creating a `noetic-devel` branch?

Not yet, before forking I need to land code cleanup changes.

dirk-thomas:
> before forking I need to land code cleanup changes.

Done in #240 and #241.

PR #236, opened 2020-01-26
restrict boost dependencies to components used

mikaelarguedas:
May be worth targeting to a noetic-devel to not impact released packages on other distributions

related to https://github.com/ros/ros_comm/pull/1871


dirk-thomas:
> May be worth targeting to a noetic-devel to not impact released packages on other distributions

Since the change only changes build and test dependencies it isn't changing the public API of the package. I don't see how it could affect downstream package. Therefore I will merge it into `melodic-devel`.

PR #244, opened 2020-03-12
Fix missing rosunit results in Python 3

mvollrath:
Sometimes rosunit does not write results to its results file.  This happens when:
* Running the test in Python 3.
* A test has failed.
* Only a few tests were run.

[Typical usage of this method by rostest](https://github.com/ros/ros_comm/blob/melodic-devel/tools/rostest/src/rostest/__init__.py#L146) never explicitly closes the file descriptor.  Python's behavior in this case is undefined and the buffered data can be lost in CPython 3.6.9 (as shipped in the melodic docker image).

By explicitly flushing we guarantee that data from this test is written before we return.

### DIFF
--- a/tools/rosunit/src/rosunit/xmlrunner.py
+++ b/tools/rosunit/src/rosunit/xmlrunner.py
@@ -265,10 +265,9 @@ class XMLTestRunner(object):
 
         time_taken = time.time() - start_time
         result.print_report(stream, time_taken, out_s, err_s)
+        stream.flush()
 
         result.print_report_text(sys.stdout, time_taken, out_s, err_s)
-
-        stream.flush()
 
         return result
 
--- a/tools/rosunit/src/rosunit/xmlrunner.py
+++ b/tools/rosunit/src/rosunit/xmlrunner.py
@@ -268,6 +268,8 @@ class XMLTestRunner(object):
 
         result.print_report_text(sys.stdout, time_taken, out_s, err_s)
 
+        stream.flush()
+
         return result
 
     def _set_path(self, path):


mvollrath:
Is this the right base branch for melodic maintenance?

dirk-thomas:
> Is this the right base branch for melodic maintenance?

The `noetic-devel` branch is the code used for ROS Noetic. But any fix which is applicable to the default branch (which is `noetic-devel`) should be merged there first. Periodically changes are being considered for backporting to older ROS distros.

So yes, this is the right branch to target.

dirk-thomas:
Regarding the proposed change: I am surprised that this should be the case. I would assume that at latest when Python exits the content will be flushed and the file handle closed.

Can you point to a bug report in Python for that behavior? Can you provide a reproducible example which demonstrates the problem?

mvollrath:
Here is an example: https://github.com/mvollrath/example_broken_rostest

When I run it with the python3 shebang:

```
$ rostest src/example_broken_rostest/test/test_flush.test 
... logging to /home/matt/.ros/log/rostest-weathertower-27833.log
[ROSUNIT] Outputting test results to /home/matt/.ros/test_results/example_broken_rostest/rostest-test_test_flush.xml
[Testcase: testtest_flush] ... ok

[ROSTEST]-----------------------------------------------------------------------


SUMMARY
 * RESULT: SUCCESS
 * TESTS: 0
 * ERRORS: 0
 * FAILURES: 0

rostest log file is in /home/matt/.ros/log/rostest-weathertower-27833.log
```

When I switch to python2 shebang:

```
$ rostest src/example_broken_rostest/test/test_flush.test 
... logging to /home/matt/.ros/log/rostest-weathertower-27789.log
[ROSUNIT] Outputting test results to /home/matt/.ros/test_results/example_broken_rostest/rostest-test_test_flush.xml
[Testcase: testtest_flush] ... ok

[ROSTEST]-----------------------------------------------------------------------

[example_broken_rostest.rosunit-test_flush/test_fail][FAILURE]------------------
x
  File "/usr/lib/python2.7/unittest/case.py", line 329, in run
    testMethod()
  File "/home/matt/catkin_ws/src/example_broken_rostest/test/test_flush.py", line 14, in test_fail
    self.fail('x')
  File "/usr/lib/python2.7/unittest/case.py", line 410, in fail
    raise self.failureException(msg)
--------------------------------------------------------------------------------

[example_broken_rostest.rosunit-test_flush/test_pass][passed]

SUMMARY
 * RESULT: FAIL
 * TESTS: 2
 * ERRORS: 0
 * FAILURES: 1

rostest log file is in /home/matt/.ros/log/rostest-weathertower-27789.log
```

mvollrath:
I've tried isolating some things in Python, but haven't found the magic yet.

mvollrath:
Same result when running `catkin_make run_tests` and `catkin test`, when the test is Python 3 no results are written and everything is green.

mvollrath:
Everything works when there are more tests, e.g. copying `test_fail` 10 times.  Probably because it's filling the first buffer.

dirk-thomas:
While I can't reproduce the behavior in a simple standalone script I can reproduce the case with the test in your test package.

The problem with the current patch is that the `XMLTestRunner` doesn't necessarily own the stream. It might have been passed in from the outside. So unconditionally closing it isn't an option.

Please see #245 and ros/ros_comm#1910 for an alternative patch.

mvollrath:
> The problem with the current patch is that the `XMLTestRunner` doesn't necessarily own the stream. It might have been passed in from the outside. So unconditionally closing it isn't an option.

Where does this patch unconditionally close the stream?

mvollrath:
Alternative patches are working for me.

dirk-thomas:
> Where does this patch unconditionally close the stream?

After I started looking into this I also tried to close the file handle. Your patch clearly doesn't do that - my mistake.

tools/rosunit/src/rosunit/xmlrunner.py
```diff
@@ -268,6 +268,8 @@ def run(self, test):
 
         result.print_report_text(sys.stdout, time_taken, out_s, err_s)
 
+        stream.flush()
```
dirk-thomas: Can you please move this line a few lines up - to be directly after the `result.print_report(stream, ...)` call.

(I would have done the change myself but you looks like you unchecked the option to allow maintainers of the target repo to make changes to your PR.)

mvollrath:
> After I started looking into this I also tried to close the file handle.

Understood, this was the first thing I tried too.

> (I would have done the change myself but you looks like you unchecked the option to allow maintainers of the target repo to make changes to your PR.)

That's odd, was not intentional.

dirk-thomas (approved):

dirk-thomas:
Thank you for the fix!

dirk-thomas:
Backported to `kinetic-devel` in 1a22b83f7f263e800503387eb6c0a4291c7a764b.

PR #245, opened 2020-03-13
close XMLTestRunner stream to ensure it is being written in Python 3

dirk-thomas:
Replaces #244.


dirk-thomas:
An updated version of #244 is sufficient - closing this one.

PR #250, opened 2020-04-06
Fix log purge on Windows

ooeygui:
Minor change to the purge code path on Windows.


dirk-thomas:
Does the same patch apply to the latest branch `noetic-devel`? If yes, please retarget that branch. If no, please explain why not.

dirk-thomas:
Closing in favor of #251.

PR #251, opened 2020-04-07
Fix log purge on Windows

ooeygui:
(Noetic port) rd is a command intrinsic of cmd.exe, not a stand alone command.


dirk-thomas:
Thanks for the fix.

seanyen:
@dirk-thomas Thanks for the merge. Can we cherry-pick it into `kinetic-devel` too, which is for `melodic`?

dirk-thomas:
Backported to `kinetic-devel` in d455dd3f4e65dd91d84fb1fed542d9528556e59e.

PR #254, opened 2020-04-12
Update rosmv function to move file and package both

Hardikrathod01:
rosmv function to support both file and package move operation


dirk-thomas:
Before the command was straight forward: it needed three arguments. In case less were provided the error message might not be great through.

With this change the command doesn't do anything when neither `-d` nor `-f` is specified (just from reading the patch, I haven't tried running this). I think this is rather intuitive. I would suggest to make the move-whole-pkg case the special case requiring a `-d` but leave the normal case to move a single file as is. Also in case the wrong number of arguments are passed a proper error message would be helpful.

Hardikrathod01:
Changes are implemented. As suggested the intuitive approach is kept as it is and option to move entire package is added.

dirk-thomas:
Please provide some example commands which you have tested the implementation with.

Hardikrathod01:
1. rosmv -d (expected output: show the help of command becuase there is no enough arguments)
2. rosmv --help (show the help)
3. rosmv <packageName> <fileName> <destination> (tested with autocompletion working)
4. rosmv -d <packageName> <destination> (tested with autocompletion working)
5. rosmv -d <packageName> (result will show the help text, because destination is not provided)
6. rosmv <packageName> <fileName> (result will show the help text because no destination is provided)

In summary if there is not enought arguments passed it will always show the help text of how to use the command) 

All the commands are tested with autocompletion.

tools/rosbash/rosbash
```diff
@@ -230,16 +230,23 @@ function rosls {
...
+        _ros_decode_path ${2} forceeval
+        pic="${rosvals[1]}${rosvals[2]}${rosvals[3]}"
+        mv -t ${3} ${pic}
+    else
+         _roscmd ${1} ${2}
+        mv ${arg} ${3}
+    fi
 }
 
+
```
dirk-thomas: Nitpick: please avoid unnecessary whitespace changes.

tools/rosbash/rosbash
```diff
@@ -1026,4 +1082,6 @@ complete -F "_roscomplete_rossrv" "rossrv"
 complete -F "_roscomplete_roscreate_pkg" "roscreate-pkg"
 complete -F "_roscomplete_roswtf" -o filenames "roswtf"
 complete -F "_roscomplete_rosconsole" "rosconsole"
-complete -F "_roscomplete_file" "rosmv"
+complete -F "_roscomplete_filemv" "rosmv"
+
+
```
dirk-thomas: Please remove added extra empty lines at the end of the file.

Also the file node bits have been changed from 644 to 755. Please revert too.

dirk-thomas (commented):
The same comments apply to all files.

tools/rosbash/rosbash
Hardikrathod01: Updated

tools/rosbash/rosbash
Hardikrathod01: Changes done to all the files

tools/rosbash/rosbash
```diff
@@ -1026,4 +1081,5 @@ complete -F "_roscomplete_rossrv" "rossrv"
 complete -F "_roscomplete_roscreate_pkg" "roscreate-pkg"
 complete -F "_roscomplete_roswtf" -o filenames "roswtf"
 complete -F "_roscomplete_rosconsole" "rosconsole"
-complete -F "_roscomplete_file" "rosmv"
+complete -F "_roscomplete_filemv" "rosmv"
+
```
dirk-thomas: Still an extra empty line.

tools/rosbash/rosbash
```diff
@@ -1004,6 +1010,55 @@ function _roscomplete_rosconsole {
...
+               COMPREPLY=($(compgen -W "$opts" -- ${arg}))
+               IFS="$_rosbash_roscomplete_search_dir_IFS"
+               unset _rosbash_roscomplete_search_dir_IFS
+            fi
+        fi
+    fi
+
+}
+
+
```
dirk-thomas: Still an extra empty line.

tools/rosbash/roszsh
```diff
@@ -797,3 +853,4 @@ compctl -x 'p[1]' -k "(md5 package packages show users)" - 'p[2]' -S '' -K "_ros
 compctl -x 'p[1]' -k "(md5 package packages show users)" - 'p[2]' -S '' -K "_roscomplete_rossrv" -- "rossrv"
 compctl -K "_roscomplete_roscreate_pkg" "roscreate-pkg"
 compctl -/g '*.(launch|test)' -x 'S[--]' -k "(--all --no-plugins --offline)" -- "roswtf"
+
```
dirk-thomas: Still an extra empty line.

tools/rosbash/rostcsh
```diff
@@ -47,3 +47,4 @@ complete rosrun \
 complete roslaunch \
    'p/1/`rospack list-names && find . -maxdepth 3 -regex ".*[xl][am][ul]n*c*h*" -printf "%P\n" `/' \
    'p@2@`rospack find $:1 | xargs -I 1 find 1 -maxdepth 3 -regex ".*[xl][am][ul]n*c*h*" -type f -printf "%f\n"`@'
+
```
dirk-thomas: Still an extra empty line.

dirk-thomas (commented):
And the permission changes are still pending.

tools/rosbash/rostcsh
Hardikrathod01: Done

tools/rosbash/roszsh
Hardikrathod01: Done

tools/rosbash/rosbash
Hardikrathod01: Done

tools/rosbash/rosbash
Hardikrathod01: I was under the impression that we leave one space at the end of the file. :)
Thank you for pointing out.

dirk-thomas: A file should end with a newline. But not with two or more.

If your editor doesn't show the already present newline I think it is time for a new one :wink:

tools/rosbash/rosfish
```diff
@@ -1052,4 +1097,5 @@ complete -x -c rossrv         -a '(_roscomplete_rossrv)'
 complete -x -c roscreate-pkg  -a '(_roscomplete_roscreate_pkg)'
 complete -x -c roswtf         -a '(_roscomplete_roswtf)'
 complete -x -c rosconsole     -a '(_roscomplete_rosconsole)'
-complete -x -c rosmv          -a '(_roscomplete_file)'
+complete -x -c rosmv          -a '(_roscomplete_filemv)'
+
```
dirk-thomas: Same here.

Hardikrathod01: Done

dirk-thomas:
Thanks for the patch.

PR #260, opened 2020-05-27
[rosbuild] fix python (urllib / open) with Python 3

k-okada:
you can check with following command:
```
download_checkmd5.py http://github.com/jsk-ros-pkg/archives/raw/master/FF-v2.3.tgz build/FF-v2.3.tgz ed33d32bdbf3f37f03c8f407209bfa47
```

### DIFF
--- a/core/rosbuild/core/rosbuild/bin/download_checkmd5.py
+++ b/core/rosbuild/core/rosbuild/bin/download_checkmd5.py
@@ -35,7 +35,7 @@
 import hashlib
 import os
 import sys
-import urllib
+import urllib.request
 from optparse import OptionParser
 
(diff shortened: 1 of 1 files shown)


dirk-thomas:
Thanks for the fix.

Please note that you shouldn't rely on `rosbuild` anymore since it has been superseded for several years already and is not being tested anymore.

k-okada:
@dirk-thomas thaks for comment, we do not use `rosbuild`, but we use `download_checkmd5.py`. May be we could move this script to other places..

k-okada:
@dirk-thomas BTW, please release new version into Noetic, when you have time. Thanks.

PR #267, opened 2020-07-17
remove encode/decode from roscreate-pkg with Python 3

roomedia:
Without the patch roscreate-pkg shows AttributeError: 'str' object has no attribute 'decode'


dirk-thomas:
Thank you for the patch.

I modified it slightly to keep explicitly encoding the content as UTF-8 and not make that subject to the default encoding: 138cb51d1690528904992201f32c6aa1c0cb6027.

Issue #268, opened 2020-09-14
rosbash does not properly quote command line arguments


SidFaber:
Arguments passed between functions within rosbash generally are not quoted which causes unexpected argument splitting and file globbing. As a result specially crafted arguments sent to some rosbash commands can result in unexpected behavior. For example:

```
roscd "roscpp forceeval b c"
```
returns the error
```
bash: [: too many arguments
```
but the current working directory is now `/opt/ros/noetic/share/roscpp`. The extra arguments `b` and `c` are ignored.

Similarly it is possible to cause unexpected behavior through specifically crafted directory names:

```
mkdir ~/tmp
cd ~/tmp
mkdir " roscpp"
mkdir forceeval
roscd "* foo"
```

This generates the error

```
bash: [: too many arguments
```
but the current working directory is now `/opt/ros/noetic/share/roscpp`.

PR #269, opened 2020-09-14
Add quoting to prevent unwanted globbing and word splitting in rosbash

SidFaber:
Unexpected arguments sent to some rosbash commands can result in unexpected behavior; for example,
```
roscd "roscpp forceeval b c"
```
prints a bash error but changes to the roscpp directory. Similar issues can happen with specially named directories.

This update addresses https://github.com/ros/ros/issues/268, and is scoped to parsing/splitting/globbing issues for arguments from the command line and local file directories only.

Signed-off-by: Sid Faber <sid.faber@canonical.com>

### DIFF
--- a/tools/rosbash/rosbash
+++ b/tools/rosbash/rosbash
@@ -45,16 +45,16 @@ function _ros_location_find {
         return 0
     fi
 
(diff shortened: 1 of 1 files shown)


kyrofa (approved):
Thanks @SidFaber, looks good to me.

dirk-thomas:
Thanks for the patch.

Issue #145, opened 2017-05-11
roscp doesn't work for installed files not in a subfolder of the manifest


dirk-thomas:
E.g. `roscp roscpp_tutorials TwoInts.h .` fails since it only considers files under `/opt/ros/lunar/share/roscpp_tutorials`.

mikaelarguedas:
Applies to `rosed` command as well

AditiPawaskar:
I want to try and solve this bug. Could anyone please guide me how to go about it?

dirk-thomas:
The files are installed in an FHS compliant location. The manifest is located in `share/<pkgname>`. The `roscp` and `rosed` commands both use [_roscmd](https://github.com/ros/ros/blob/782d333f26f6f6f020412b3f56669ecace690d81/tools/rosbash/rosbash#L233) which only looks in a few possible locations. In order to work for other locations the function need to be extended or a new function being written which finds files in other locations.

PREM1980:
@dirk-thomas Planning to submit a PR on this issue. Want to make sure no one else is working on it.

dirk-thomas:
> Want to make sure no one else is working on it.

Nobody working on this afaik. Thank you, looking forward for your pull request.

PREM1980:
@dirk-thomas : Thanks.

I tried to execute the below command after installing lunar but get the following error - 

m-ubu1604:/opt/ros/lunar/share/roscpp_tutorials$ roscp roscpp_tutorials TwoInts.h .
That file does not exist in that package.
cp: missing destination file operand after '.'
Try 'cp --help' for more information.

What is the expected result?  or can I know what is the  problem?

dirk-thomas:
> That file does not exist in that package.

This error message indicates that the script didn't find the source file.

> cp: missing destination file operand after '.'
> Try 'cp --help' for more information.

This is probably a follow up error because the source was not found the destination `.` is the only argument to `cp` which is insufficient.

> What is the expected result? or can I know what is the problem?

The expected behavior is that the file file `TwoInts.h` in the package `roscpp_tutorials` is copied to the destination `.`. You will have to debug the logic in the `roscp` function to figure what went wrong and how to fix it.

PREM1980:
ok, I don't see the TwoInts.h within the roscpp_tutorials directory..

m-ubu1604:/opt/ros/lunar/share/roscpp_tutorials$ tree
.
├── cmake
│   ├── roscpp_tutorialsConfig.cmake
[...]

PR #274, opened 2020-10-09
Added new function roscpd to address issue #145

dharapatel1596:
A new ros function "roscpd" is added whose functionalities are as follows:

- Copy a directory (inside or outside catkin workspace) **recursively** with all files and folders to a target location.
- Copy a single file (inside or outside catkin workspace) to a target location. 

This functionality is added to address the issue #145.

### DIFF
--- a/tools/rosbash/rosbash
+++ b/tools/rosbash/rosbash
@@ -299,6 +299,14 @@ function roscp {
     cp "${arg}" "${3}"
 }
 
+function roscpd {
+    if [[ $1 = "--help" ]] || [[ $# -ne 2 ]]; then
+        echo -e "usage: roscpd [directory/filepath] [target]\n\nCopy a (directory/file) to target location."
+        return 0
+    fi
+    cp -r ${1} ${2}
+}
+
 function roscat {
     local arg
     if [[ $1 = "--help" ]] || [[ $# -ne 2 ]]; then
@@ -1058,6 +1066,14 @@ function _roscompletemv_search_dir {
 
 }
 
+function _roscomplete_roscpd {
+
+  local arg=${COMP_WORDS[COMP_CWORD]}
+  local IFS=$'\n'
+  COMPREPLY=()
+  COMPREPLY=( $( compgen -f -- ${arg} ) )
+}
+
 complete -F "_roscomplete_sub_dir" -o "nospace" "roscd"
 complete -F "_roscomplete_sub_dir" -o "nospace" "rospd"
 complete -F "_roscomplete_sub_dir" -o "nospace" "rosls"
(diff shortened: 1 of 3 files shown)


hidmic (commented):
I'm not sure I follow how this patch addresses #145. 

Correct me if I'm wrong, but adapting the repro given there:

```
roscpd roscpp_tutorials TwoInts.h .
```

would result in the third argument being ignored and an attempt to recursively copy some local `roscpp_tutorials` directory into a `TwoInts.h` directory.

I think you may want to revisit the discussion on #145.

dharapatel1596:
So according to what I followed from the discussion is that: the command 'roscp' looks for the file only in the specified package and if the file is not present there it will result into one or the other error.

So, I tried to introduce a new command 'roscpd' which can now look for the files and directories both **inside catkin workspace** or also **outside catkin workspace**. 

If you look at the help of the command 'roscpd', it does not take three arguments. 
So testing the command with:

`roscpd roscpp_tutorials TwoInts.h .`

will result into showing the help of the command.

With this command now, package name is not required. I tried to keep the command short. 

Let me present an example of testing above command:

`roscpd src/roscpp_tutorial/<filename> .`

will copy the **provided file** to the **current location**.

`roscpd src/roscpp_tutorial/ .` 

will copy the package **roscpp_tutorial** to the **current directory**. 

So, instead of restricting the search into a package, I made it like, looking for the file into a directory path provided. This directory path can be **relative** as well as **absolute** from the current location. 

Hence, user now has an option to copy a file even if it is not located into the package and hence, enabling the user to copy it from anywhere in the system to the target location. 

I think this resolves the issue. 

Let me know what you think.

hidmic:
Hmm, `roscp`'s raison d'être is to not have to fiddle with package paths. As-is, `roscpd` is just an alias for `cp -r`.

IIUC #145 is about adding other search paths to `roscp` so it can find files that are outside a package's share directory (which may or may not be possible depending on the install space layout). PTAL @ #145.

dharapatel1596:
Okay. So command `roscpd` is not resolving the Issue #145. But can we go ahead with introducing `roscpd` as the new command, for copying the directory/file from anywhere to the target location?

hidmic:
> But can we go ahead with introducing roscpd as the new command, for copying the directory/file from anywhere to the target location?

I'm reluctant to introduce an alias for `cp`. Why not just use `cp`? But I'll let @jacobperron chime in too.

jacobperron:
I have similar feelings as @hidmic. `ros*` commands exist to operate on ROS-aware install/build prefixes. As pointed out, the proposed `roscpd` is simply an alias for `cp -r`. I.e. What advantage does `roscpd` have over running `cp -r`?

dharapatel1596:
As there is not much advantage of `roscpd` over `cp -r` .
It is more or less same. I was trying to give abstraction with ROS command. I agree with your point too.

jacobperron:
> As there is not much advantage of roscpd over cp -r .
It is more or less same. I was trying to give abstraction with ROS command. I agree with your point too.

On the premise that this PR is not adding much value, I'm inclined to close this this ticket.

PR #278, opened 2021-03-04
Keep rosed from running if the file doesn't exist in the package

timonegk:
Same as abb02a22d4c67638dcc95a80241ade95658a3b02, but for zsh.

### DIFF
--- a/tools/rosbash/rosbash
+++ b/tools/rosbash/rosbash
@@ -241,7 +241,8 @@ function rosmv {
         pic="${rosvals[1]}${rosvals[2]}${rosvals[3]}"
         mv -t "${3}" "${pic}"
     else
-         _roscmd "${1}" "${2}"
+        _roscmd "${1}" "${2}"
+        [ $? -eq 1 ] && return 1
         mv "${arg}" "${3}"
     fi
 }
@@ -280,12 +281,11 @@ function rosed {
        return 0
     fi
     _roscmd "${1}" "${2}"
-    if [[ -n ${arg} ]]; then
-        if [[ -z $EDITOR ]]; then
-            vim "${arg}"
-        else
-            $EDITOR "${arg}"
-        fi
+    [ $? -eq 1 ] && return 1
+    if [[ -z $EDITOR ]]; then
+        vim "${arg}"
+    else
+        $EDITOR "${arg}"
     fi
 }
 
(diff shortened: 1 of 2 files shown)


hidmic (approved):
LGTM, @sloretz any thoughts?

jacobperron (commented):
Sounds reasonable to keep the behavior consistent, though I'm not sure if we want to break behavior mid-distro.

hidmic:
> though I'm not sure if we want to break behavior mid-distro.

Fair point. I don't have a strong opinion. It is somewhat surprising for the same tool to behave differently for each shell though.

timonegk:
I just saw the check is done for `roscat` (using `$?` and `return`) and thought that would be better for the other tools, too.

timonegk:
Also, I don't think that any of the changes is actually a breaking change. `rosed` with an unknown package name currently just opens an empty editor without any file selected, I can't imagine any use case for that. Also, `rosed` is not a tool that can be used in scripts.
`roscp` and `rosmv` currently fail, but with the following message:
```
Couldn't find package [foo]
cp: missing destination file operand after '/path/to/destination'
Try 'cp --help' for more information.
```
(and the same for `mv`)
This is changed to just the first line (`Couldn't find package [foo]`), which is less confusing.

timonegk:
While we're at it, I just applied the unification to bash as well.

hidmic (approved):
As I said above, I'm onboard with these changes. Regarding:

> rosed with an unknown package name currently just opens an empty editor without any file selected, I can't imagine any use case for that. Also, rosed is not a tool that can be used in scripts.

@timonegk why not adjusting `rosed` to skip that `_roscmd` call if `$#` is below 1 and just open the editor? That way we don't break any use case (no matter how odd it may be).

timonegk:
Sure, I updated the files.
Before you ask: I removed the quotes in rosbash because then, the editor can be called when $arg is empty. When it is called with "", the behavior is different. This might break when there are spaces in the file name, but that already breaks in _roscmd and should be fixed there, which is a different issue.

hidmic:
>  I removed the quotes in rosbash because then, the editor can be called when $arg is empty. When it is called with "", the behavior is different. This might break when there are spaces in the file name, but that already breaks in _roscmd and should be fixed there, which is a different issue.

Ah, that's not ideal, but OK. @jacobperron ?

hidmic:
@jacobperron any thoughts against merging?

jacobperron:
@hidmic Sounds okay to me. Sorry for the delay.

PR #262, opened 2020-06-15
fix for sourcing rosfish

iPaddy:
Fixes #261 
Fixed the script such that it sources in fish.
Functionality should match the intentions of the bash part.

PR #279, opened 2021-03-16
Fix variable name (from $arg to $argv) in rosfish

eduidl:
Completion does not work well due to this bug.

ref: https://github.com/ros/ros/pull/262


jacobperron (approved):
Thanks for the fix!

PR #280, opened 2021-04-12
roszsh: fix rosmv completion

timonegk:
The current completion for `rosmv` and zsh is just a copy of the bash completion. It does not work and instead sets the `$path` variable, which is used internally by zsh instead of `$PATH`, so that the current shell cannot be used further. This PR fixes the completion by calling the functions that already exist (`_roscomplete` and `_roscomplete_file`) for package and package file completion and using the zsh-internal file completion instead of rewriting it.

### DIFF
--- a/tools/rosbash/roszsh
+++ b/tools/rosbash/roszsh
@@ -787,51 +787,15 @@ function _roscomplete_roscreate_pkg {
 }
 
 function _roscomplete_filemv {
-    _roscompletemv_search_dir "-type f ! -regex .*[.][oa]$"
-}
-
-function _roscompletemv_search_dir {
-    local arg opts rosvals
-    COMPREPLY=()
-    arg="${COMP_WORDS[COMP_CWORD]}"
-    option="$(echo -e "${COMP_WORDS[1]}" | tr -d '[:space:]')"
-    if [[ $COMP_CWORD == 2 && ( $option == "-d" ) ]] || [[ $COMP_CWORD == 1 ]]; then
+    local words
+    words=(${=BUFFER})
+    if [[ $CURRENT == 3 && ( $words[2] == "-d" ) ]] || [[ $CURRENT == 2 ]]; then
         # complete packages
-        _roscomplete_pkg "${arg}"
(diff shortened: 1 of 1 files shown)


tools/rosbash/roszsh
```diff
@@ -787,59 +787,23 @@ function _roscomplete_roscreate_pkg {
...
-               IFS=$'\n'
-               COMPREPLY=($(compgen -W "$opts" -- ${arg}))
-               IFS="$_rosbash_roscomplete_search_dir_IFS"
-               unset _rosbash_roscomplete_search_dir_IFS
-            fi
-        fi
+        _roscomplete
+    elif [[ $CURRENT == 3  && ( $words[1] != "-d" ) ]]; then
+        # complete files within package
+        _roscomplete_file
```
hidmic, line 797: @timonegk I'm not well versed in `compctl` on `zsh`, but isn't this patch removing the functionality in the last clause completely?

timonegk: Do you mean the else case that does the normal file completion? That is handled by zsh, as specified in the `compctl` command below. There, `-f` means that file names and directories are completed. With `-x`, a different completion is be specified for some cases, in this case for `p[1,2]` which means the first and second word of the arguments. `-K "_roscomplete_filemv"` means that in this case, the `_roscomplete_filemv` function is used to get the completion. `--` ends the arguments to `-x`, `rosmv` specifies the command for which the completion is applied.
Therefore, the `_roscomplete_filemv` function is only used for the first two arguments, which is either a package and a file, or `-d` and a package. In both cases, the third argument is a filename or path that can be completed with the normal `-f`. The overly complex code that was previously used is completely unnecessary (and did not work anyway).

hidmic: Thanks for the explanation!

hidmic (approved):

hidmic:
@jacobperron any thoughts against merging?

jacobperron (approved):
LGTM

jacobperron:
@timonegk Thanks for the fix!

PR #318, opened 2023-08-24
rosrun: Do not consider binaries from directories with CATKIN_IGNORE

peci1:
Some IDEs (CLion) create build directories under package source folders whose name is not `build` (CLion makes `cmake-build-debug` etc.). These directories are properly initialized with catkin, so a `CATKIN_IGNORE` is put into them. However, `rosrun` still finds executables in these folders and in case of Python scripts, this results in a non-unique file to be run, which in turn results in the need of user input, which in turn makes it impossible to run these executables from launch files.

This PR fixes it and no executables under an ignored directory are considered.

However, I'm not sure if running ignored executables is not actually a valid use-case. If so, could the solution be that this pruning is only done in case there is exactly one non-ignored and one or more ignored executables?

### DIFF
--- a/tools/rosbash/scripts/rosrun
+++ b/tools/rosbash/scripts/rosrun
@@ -21,7 +21,7 @@ function parent_find() {
   test -e "$dir/$file" && readlink -f "$dir" && return 0
   [ '/' = "$dir" ] && return 1
 
-  parent-find "$file" "$(readlink -f "$dir/..")"
+  parent_find "$file" "$(readlink -f "$dir/..")"
 }
 
 while [ $args == 1 ]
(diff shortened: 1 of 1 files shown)


sloretz:
Thank you for the PR!

[ROS Noetic will reach end-of-life on May 31st, 2025](https://discourse.ros.org/t/ros-noetic-end-of-life-may-31-2025/43160). Every change comes with a risk of introducing regressions, and there isn't much time left to fix them. To make sure this PR doesn't introduce any regressions please:

* Describe how you tested this change
* Recruit at least one more person to review this PR and try it out on their system

sloretz:
ROS 1 is end-of-life (EOL) as of today, May 31st 2025. I am archiving this repository because:

* it only supports ROS 1
* it isn't needed anymore in ROS 2

[If you still rely on ROS 1, read this page to learn about your options.](https://bit.ly/NoeticEOL)

